PR #45, opened 2016-04-19
Add find me button for location enabled

### LINKED ISSUE #29: setMyLocationEnabled(true) should also show find me button
If we want to model the prevailing map behavior then showing a UI component to center on the users current location (and start tracking updates) will also be enabled when `setMyLocationEnabled(boolean)` is set to true.

https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.html#setMyLocationEnabled(boolean)

sarahlensing:
Closes #29


library/src/main/java/com/mapzen/android/TangramMap.java
```diff
@@ -0,0 +1,27 @@
+package com.mapzen.android;
+
+import android.content.Context;
+import android.util.AttributeSet;
+
+/**
+ * Entry point for interation with Tangram map. Wrapped inside of {@link MapView}.
```
ecgreb: interaction + integration = interation ?

;)

library/src/main/res/layout/view_map.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ecgreb: Can this be a `<merge>` tag to reduce layout depth?

Right now `MapView` extends `FrameLayout` so we have a `RelativeLayout` inside a `FrameLayout`. Perhaps `MapView` should instead extend `RelativeLayout` and then we use a `<merge>` tag here to wrap the map and find me button?

library/src/main/java/com/mapzen/android/TangramMap.java
sarahlensing: :) ...updated to interaction!

library/src/main/res/layout/view_map.xml
sarahlensing: Indeed! Nice catch, updated.

PR #47, opened 2016-04-20
Add CONTRIBUTING.md

sarahlensing:


ecgreb:
I'm not sure we want to use the same `CONTRIBUTING` template from Eraser Map here. That one seems more geared to a consumer app.

For the SDK I was thinking something more along these lines: https://github.com/square/android-times-square/blob/master/CONTRIBUTING.md

sarahlensing:
That's definitely more relevant. Updated.

PR #49, opened 2016-04-20
Adds MapzenMap

ecgreb:
Adds `MapzenMap` wrapper for `MapController`.

TODO: Make `OverlayManager` a member of `MapzenMap` to provide a single class API for all common map operations including markers and overlays. Add wrapper methods for common `MapController` methods. Add ability to switch between bundled stylesheets.

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -45,4 +45,5 @@ dependencies {
   }
   compile 'com.android.support:appcompat-v7:23.2.1'
   compile 'com.android.support:design:23.2.1'
+  compile 'com.mapzen.tangram:tangram:0.2-SNAPSHOT'
 }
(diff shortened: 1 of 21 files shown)


sample/build.gradle
```diff
@@ -45,4 +45,5 @@ dependencies {
   }
   compile 'com.android.support:appcompat-v7:23.2.1'
   compile 'com.android.support:design:23.2.1'
+  compile 'com.mapzen.tangram:tangram:0.2-SNAPSHOT'
```
sarahlensing: I don't think we need this because we are transitively including dependencies from the library project

ecgreb: I re-added this b/c I was running into issues with IntelliJ/Android Studio not resolving Tangram classes in the sample app. Command line build would work fine but the IDE was full of red. Is it working normally for you?

sarahlensing: It's working normally on my end. Hrm.

PR #83, opened 2016-05-09
Support loading custom stylesheet

### LINKED ISSUE #82: Add ability to set custom stylesheet

sarahlensing:
Closes #82


library/src/main/java/com/mapzen/android/MapView.java
```diff
@@ -126,6 +126,28 @@ public void getMapAsync(@NonNull String key,
...
+        mapInitializer.init(this, key, sceneFile, callback);
+    }
+
+    /**
+     * Load map asynchronously using given API key.
+     *
+     * @param callback listener to be invoked when map is initialized and ready to use.
+     * @param sceneFile scene file that should be used
+     */
+    public void getMapAsync(@NonNull OnMapReadyCallback callback, @NonNull String sceneFile) {
```
ecgreb: We should probably switch the order of the params and make `callback` last here to keep consistent with the other versions of this method.

sarahlensing: We cant though because we already have a method with the same signature for the keys. It seems like we have a couple options though- 1. make all the getMapAsync methods which take a scene file have the scene file as the last param, 2. change the name of the method where we pass a scene file to something like getCustomMapAsync. Do you have a preference? I think I have a slight preference for changing the method name

sample/src/main/java/com/mapzen/android/sample/CustomStylesheetActivity.java
```diff
@@ -0,0 +1,77 @@
+package com.mapzen.android.sample;
+
+import com.mapzen.android.MapFragment;
+import com.mapzen.android.MapzenMap;
+import com.mapzen.android.OnMapReadyCallback;
+import com.mapzen.android.model.BubbleWrapStyle;
```
ecgreb: Unused imports

PR #86, opened 2016-05-12
Adopt Square Android code style?

ecgreb:
https://github.com/square/java-code-styles

Similar to what we have now except:

* 2-space indent
* Inline annotations
* XML layout closing tag bracket wraps to next line

Thoughts?


sarahlensing:
Artisanal code styling :P

I'm okay with these changes, it should give us a little extra line space for code

msmollin:
+1 for Artisanal Code Styling.

PR #88, opened 2016-05-12
Add routing overlay methods

sarahlensing:
Adds support for common routing actions in preparation for the EM migration

### DIFF
--- a/library/src/main/java/com/mapzen/android/MapzenMap.java
+++ b/library/src/main/java/com/mapzen/android/MapzenMap.java
@@ -460,12 +460,18 @@ public class MapzenMap {
      * Draws search results on the map. All pins will be inactive except for the one at the active
      * index supplied.
      * @param points
-     * @param active
+     * @param activeIndexes
      */
-    public void drawSearchResults(List<LngLat> points, int active) {
+    public void drawSearchResults(List<LngLat> points, int... activeIndexes) {
         int index = 0;
         for (LngLat point : points) {
-            boolean status = active == index;
+            boolean status = false;
+            for (int i = 0; i < activeIndexes.length; i++) {
+                if (activeIndexes[i] == index) {
+                    status = true;
+                    break;
+                }
+            }
             overlayManager.drawSearchResult(point, status, index);
             index++;
         }
--- a/library/src/main/java/com/mapzen/android/MapzenMap.java
+++ b/library/src/main/java/com/mapzen/android/MapzenMap.java
@@ -10,8 +10,8 @@ import com.mapzen.tangram.MapController;
 import com.mapzen.tangram.MapData;
 import com.mapzen.tangram.TouchInput;
 
-
 import java.util.HashMap;
+import java.util.List;
 
 /**
  * This is the main class of the Mapzen Android API and is the entry point for all methods related
@@ -20,6 +20,7 @@ import java.util.HashMap;
  * {@link MapView#getMapAsync(OnMapReadyCallback)}.
  */
 public class MapzenMap {
+
     private final MapController mapController;
     private final OverlayManager overlayManager;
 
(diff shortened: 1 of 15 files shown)


library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -394,4 +395,129 @@ public boolean isSimultaneousGestureAllowed(TouchInput.Gestures first,
...
+     * @param point
+     */
+    public void drawDroppedPin(LngLat point) {
+        overlayManager.drawDroppedPin(point);
+    }
+
+    /**
+     * Clears the dropped pin from the map.
+     */
+    public void clearDroppedPin() {
```
ecgreb: Should this method be renamed to `clearDroppedPins()` since it clears all dropped pins from the map?

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -394,4 +395,129 @@ public boolean isSimultaneousGestureAllowed(TouchInput.Gestures first,
...
+            }
+            overlayManager.drawSearchResult(point, status, index);
+            index++;
+        }
+    }
+
+    /**
+     * Clears search result from the map.
+     */
+    public void clearSearchResult() {
```
ecgreb: I understand why we have both `clearSearchResult()` and 'clearSearchResults()` for completeness but given they have the same implementation perhaps we can delete one to avoid confusion.

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -394,4 +395,129 @@ public boolean isSimultaneousGestureAllowed(TouchInput.Gestures first,
...
+     */
+    public void clearSearchResults() {
+        overlayManager.clearSearchResult();
+    }
+
+    /**
+     * Draws route pin at the point supplied.
+     * @param point
+     */
+    public void drawRoutePin(LngLat point) {
```
ecgreb: What do you think about renaming `drawRoutePin(...)` to `drawRouteMarker(...)` or `drawRouteLocationMarker(...)` to avoid confusion with start and end pins which are drawn using `drawRoutePins(...)`?

library/src/main/java/com/mapzen/android/OverlayManager.java
```diff
@@ -183,6 +207,146 @@ public MapData addMarker(Marker marker) {
...
+    public void drawRoutePins(LngLat start, LngLat end) {
+        if (startPinData == null) {
+            startPinData = mapController.addDataLayer(NAME_START_PIN);
+        }
+        if (endPinData == null) {
+            endPinData = mapController.addDataLayer(NAME_END_PIN);
+        }
+        startPinData.addPoint(start, null);
+        endPinData.addPoint(end, null);
+        mapController.requestRender();
```
ecgreb: Need to verify with @blair1618 or @tallytalwar but I don't think calling `mapController.requestRender()` is required any longer when adding client data to the map. I tried commenting these out and everything seems to still work.

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: Updated.

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: Rm'd clearSearchResult()

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: I like that suggestion- lets go with drawRouteLocationMarker() for clarity

library/src/main/java/com/mapzen/android/OverlayManager.java
sarahlensing: Hmm we might need clarification...  tried commenting them out but when I do so I don't see the pins. I'm testing on an S6 edge.

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -394,4 +395,122 @@ public boolean isSimultaneousGestureAllowed(TouchInput.Gestures first,
...
+     * @param point
+     */
+    public void drawRouteLocationMarker(LngLat point) {
+        overlayManager.drawRouteLocationMarker(point);
+    }
+
+    /**
+     * Clears route pin from the map.
+     */
+    public void clearRoutePin() {
```
ecgreb: New method names feel better. We should probably also update this to `clearRouteLocationMarker(...)` for consistency.

library/src/main/java/com/mapzen/android/OverlayManager.java
ecgreb: We can always merge as is and remove these later if it turns out they are no longer needed. Its good for the Tangram team to know the behavior is slightly different between devices too. I am testing on the Nextbit Robin (Android 6.0.1)

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: Updated!

PR #93, opened 2016-05-17
namespace resources, add find me external click listener, ability to …

sarahlensing:
…center map programmatically, hide location marker when setMyLocationEnabled(false), don’t require mapzen.xml for keys set in code, expose find me btn and attribution text

### DIFF
--- a/library/src/main/java/com/mapzen/android/MapView.java
+++ b/library/src/main/java/com/mapzen/android/MapView.java
@@ -27,6 +27,7 @@ public class MapView extends RelativeLayout {
 
     TangramMapView tangramMapView;
     ImageButton findMe;
+    TextView attribution;
 
     /**
      * Create new instance.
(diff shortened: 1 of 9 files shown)


ecgreb:
I spent some time trying to figure out how we can test the external find me click listener and here is what I came up with. Its pretty damn verbose 😢 but a lot of this could be moved into `setup()` or other helper methods.

We are going to need to find a way to test things like this in general anyway especially if we're not going to be using Robolectric. So I kind of see this as laying the groundwork for future unit tests that touch Android framework components.

```java
    @Test
    public void setFindMeOnClickListener_shouldInvokeListenerOnButtonClick() throws Exception {
        MapView mapView = mock(MapView.class);
        MapController mapController = mock(TestMapController.class);
        LostApiClient lostApiClient = mock(LostApiClient.class);
        LocationServices.FusedLocationApi = mock(FusedLocationProviderApiImpl.class);

        TestOnClickListener listener = new TestOnClickListener();
        TestButton testFindMeButton = new TestButton(null);
        OverlayManager overlayManager = new OverlayManager(mapView, mapController, lostApiClient);

        when(mapView.showFindMe()).thenReturn(testFindMeButton);
        overlayManager.setFindMeOnClickListener(listener);
        overlayManager.setMyLocationEnabled(true);
        testFindMeButton.performClick();
        assertThat(listener.click).isTrue();
    }

    private class TestOnClickListener implements View.OnClickListener {
        private boolean click = false;

        @Override public void onClick(View v) {
            click = true;
        }
    }

    private class TestButton extends ImageButton {
        private OnClickListener listener;

        public TestButton(Context context) {
            super(context);
        }

        @Override
        public void setOnClickListener(OnClickListener l) {
            listener = l;
        }

        @Override
        public boolean performClick() {
            listener.onClick(this);
            return true;
        }
    }
```

sarahlensing:
Really nice! I've integrated it into our suite.

PR #95, opened 2016-05-19
Support scene updates

sarahlensing:
Adds MapzenMap methods:
- MapzenMap#setFeaturePickListener
- MapzenMap#setViewCompleteListener
- MapzenMap#queueEvent
- MapzenMap#setDebugFlag
- MapzenMap#queueSceneUpdate
- MapzenMap#applySceneUpdates


library/src/main/java/com/mapzen/android/model/DebugFlag.java
```diff
@@ -0,0 +1,14 @@
+package com.mapzen.android.model;
+
+/**
+ * Debug flags available on the map.
+ */
+public enum DebugFlag {
```
ecgreb: Given these debug flags are used specifically for Tangram development I'm not sure they belong in the SDK?

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -280,16 +309,30 @@ public LngLat coordinatesAtScreenPosition(double screenX, double screenY) {
     /**
      * Set tap responder for tap gestures on map.
      */
-    public void setTapResponder(final TouchInput.TapResponder tapResponder) {
-        this.tapResponder = tapResponder;
-        mapController.setTapResponder(this.tapResponder);
+    public void addTapResponder(final TouchInput.TapResponder tapResponder) {
```
ecgreb: I'm still not sure supporting multiple gesture listeners of the same type is worth the added complexity. If we decide to go this route we will need to give the developer a way to remove a single listener as well as to clear all.

For comparison, Google Maps only allows a single listener of each type. And clearing the listener is achieved by calling the same method and passing `null` as the listener param.

https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.html#setOnMapClickListener(com.google.android.gms.maps.GoogleMap.OnMapClickListener)

sarahlensing: Changed.

I had changed this so that setting a `FeaturePickListener` could encapsulate calling `pickFeature` but realize I can just set an internal `TapResponder` and have it manage a single external responder plus the call to `pickFeature`. This will keep our api on par with GoogleMap's!

library/src/main/java/com/mapzen/android/model/DebugFlag.java
sarahlensing: Removed.

PR #103, opened 2016-05-31
Prefix all resources with mz_

sarahlensing:


ecgreb:
Should we prefix individual strings and colors with `mz_` too?

sarahlensing:
Yes, we should- good call.

Prefixed.

PR #112, opened 2016-06-06
Set httphandler in MapzenRouter constructor

sarahlensing:
- Migrates `MapzenRouter` to use on-the-road's new `HttpHandler` when setting api key

Depends on: https://github.com/mapzen/on-the-road/pull/73

### DIFF
--- a/library/src/main/java/com/mapzen/android/MapzenRouter.java
+++ b/library/src/main/java/com/mapzen/android/MapzenRouter.java
@@ -1,6 +1,5 @@
 package com.mapzen.android;
 
-import com.mapzen.valhalla.HttpHandler;
 import com.mapzen.valhalla.RouteCallback;
 import com.mapzen.valhalla.Router;
 import com.mapzen.valhalla.ValhallaRouter;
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/mapzen/android/MapzenRouter.java
```diff
@@ -20,51 +20,39 @@
...
-    initializeRouterKey();
-    initializeRouterUrl();
+  public MapzenRouter(String apiKey) {
+    TurnByTurnHttpHandler httpHandler = new TurnByTurnHttpHandler();
+    httpHandler.setApiKey(apiKey);
+    internalRouter.setHttpHandler(httpHandler);
   }
 
-  private void initializeRouterKey() {
+  private String initializeRouterKey(Context context) {
```
ecgreb: What do you think about moving the routing API key initialization into `AndroidModule` so the `TurnByTurnHttpHandler` can be injected with Dagger? This is how we are currently creating and using the `TileHttpHandler`?

https://github.com/mapzen/android/blob/master/library/src/main/java/com/mapzen/android/dagger/AndroidModule.java#L51

This way all the API key initialization would be in the same place. This would also reduce the need for reflection in the tests.

sarahlensing: I think that is a good idea- let me make that change and repush

PR #141, opened 2016-07-07
WIP: Test transit route line

nvkelso:
Creating a real PR out of this so we can comment on line numbers.

### DIFF
--- a/library/src/main/java/com/mapzen/android/OverlayManager.java
+++ b/library/src/main/java/com/mapzen/android/OverlayManager.java
@@ -35,11 +35,13 @@ public class OverlayManager implements TouchInput.PanResponder {
   private static final String NAME_POLYGON = "mz_default_polygon";
   private static final String NAME_MARKER = "mz_default_point";
   private static final String NAME_START_PIN = "mz_route_start";
-  private static final String NAME_END_PIN = "mz_route_stop";
+  private static final String NAME_END_PIN = "mz_route_destination";
   private static final String NAME_DROPPED_PIN = "mz_dropped_pin";
   private static final String NAME_SEARCH_RESULT_PIN = "mz_search_result";
   private static final String NAME_ROUTE_PIN = "mz_route_location";
   private static final String NAME_ROUTE_LINE = "mz_route_line";
+  private static final String NAME_TRANSIT_ROUTE_LINE = "mz_route_line_transit";
+  private static final String NAME_STATION_ICON = "mz_route_transit_stop";
 
   private static final String PROP_STATE = "state";
   private static final String PROP_STATE_ACTIVE = "active";
@@ -95,6 +97,8 @@ public class OverlayManager implements TouchInput.PanResponder {
   private static MapData searchResultPinData;
   private static MapData routePinData;
   private static MapData routeLineData;
+  private static MapData transitRouteLineData;
+  private static MapData stationIconData;
 
   /**
    * Create a new {@link OverlayManager} object for handling functionality between map and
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/mapzen/android/OverlayManager.java
```diff
@@ -35,11 +35,13 @@
   private static final String NAME_POLYGON = "mz_default_polygon";
   private static final String NAME_MARKER = "mz_default_point";
   private static final String NAME_START_PIN = "mz_route_start";
-  private static final String NAME_END_PIN = "mz_route_stop";
+  private static final String NAME_END_PIN = "mz_route_destination";
```
nvkelso: This will be a backwards incompatible change, which is fine but needs to be coordinated with the stylesheet (scene file). Currently it's expecting `mz_route_stop` which was the old name. I think it is a good idea to change this in the app, too like you propose, so I'll push another change to the scene file for this.

nvkelso: (I had created an alias for it in the scene file yesterday, but kept the original source the same to ensure backwards compatibility with the app's interface. Sorry that was confusing!)

library/src/main/java/com/mapzen/android/OverlayManager.java
```diff
@@ -35,11 +35,13 @@
...
   private static final String NAME_DROPPED_PIN = "mz_dropped_pin";
   private static final String NAME_SEARCH_RESULT_PIN = "mz_search_result";
   private static final String NAME_ROUTE_PIN = "mz_route_location";
   private static final String NAME_ROUTE_LINE = "mz_route_line";
+  private static final String NAME_TRANSIT_ROUTE_LINE = "mz_route_line_transit";
```
nvkelso: Same deal for these.

library/src/main/java/com/mapzen/android/OverlayManager.java
```diff
@@ -334,6 +338,36 @@ public void clearRouteLine() {
...
+   * Draws route line on the map for the points supplied.
+   */
+  public void drawTransitRouteLine(List<LngLat> points, String colorHex) {
+    if (transitRouteLineData == null) {
+      transitRouteLineData = mapController.addDataLayer(NAME_TRANSIT_ROUTE_LINE);
+    }
+    HashMap<String, String> properties = new HashMap<>();
+    properties.put(PROP_TYPE, PROP_LINE);
+
+    mapController.queueSceneUpdate("layers.mz_route_line_transit.draw.ux-transit-line-overlay.color",
```
nvkelso: You should probably remove this line as my scene file code is already data driven, shouldn't need you mucking with updating the scene.

sample/src/main/java/com/mapzen/android/sample/RouteLineActivity.java
```diff
@@ -54,6 +54,10 @@
             return false;
           }
         });
+        ArrayList<LngLat> testPoints = new ArrayList();
+        testPoints.add(new LngLat(-122.39353246246766, 37.78662344689961));
+        testPoints.add(new LngLat(-122.39309926415683, 37.791273135641994));
+        map.drawTransitRouteLine(testPoints, "#ff0000");
```
nvkelso: as long as this creates a data attribute (property) like `color: '#ff0000'` on a line geometry then my style will work (rather than relying on a scene file update).

library/src/main/java/com/mapzen/android/OverlayManager.java
```diff
@@ -334,6 +338,36 @@ public void clearRouteLine() {
...
+        "'"+colorHex+"'");
+    mapController.applySceneUpdates();
```
nvkelso: Or this line, most likely.

library/src/main/assets/styles/bubble-wrap
```diff
@@ -1 +1 @@
-Subproject commit 1e100fbc0a9364de188ad35738853a7c57dc1f58
+Subproject commit 63b205e9e530045b6f3928e34b1caf68c5bc9510
```
nvkelso: Updated SHA is `e9172d25ed9d5aff90c6e94f09b34e4f7755e505`

library/src/main/java/com/mapzen/android/OverlayManager.java
nvkelso: Fixed in e9172d25ed9d5aff90c6e94f09b34e4f7755e505.

library/src/main/java/com/mapzen/android/OverlayManager.java
nvkelso: Fixed in https://github.com/tangrams/bubble-wrap/commit/e9172d25ed9d5aff90c6e94f09b34e4f7755e505.

library/src/main/java/com/mapzen/android/OverlayManager.java
sarahlensing: On a related note, we will need to update cinnabar and refill too. I'll do that and then push

sample/src/main/java/com/mapzen/android/sample/RouteLineActivity.java
sarahlensing: It now does!

sarahlensing:
Closing as no longer needed

PR #146, opened 2016-07-18
Add Pelias to SDK and include example Activity

### LINKED ISSUE #145: add Pelias to SDK

sarahlensing:
- Adds `MapzenSearch` to SDK for interaction with `Pelias` and configuration of related api key
- Adds `MapzenSearchViewActivity` to demonstrate use of `MapzenSearch` using `PeliasSearchView` and `MapzenMap`

Depends on: https://github.com/pelias/pelias-android-sdk/pull/44

Closes #145

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -85,6 +85,7 @@ dependencies {
   compile "com.mapzen.tangram:tangram:$tangram_version"
   compile 'com.mapzen.android:lost:1.1.1'
   compile 'com.mapzen:on-the-road:0.8.4'
+  compile('com.mapzen.android:pelias-android-sdk:0.7.2-SNAPSHOT')
   compile 'com.google.dagger:dagger:2.0'
   compile 'com.google.dagger:dagger-compiler:2.0'
   compile 'javax.annotation:javax.annotation-api:1.2'
--- a/library/src/main/java/com/mapzen/android/dagger/AndroidModule.java
+++ b/library/src/main/java/com/mapzen/android/dagger/AndroidModule.java
@@ -1,5 +1,7 @@
 package com.mapzen.android.dagger;
 
+import com.mapzen.android.ApiKeyConstants;
+import com.mapzen.android.SearchInitializer;
 import com.mapzen.android.TileHttpHandler;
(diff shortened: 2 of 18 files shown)


sample/src/main/res/layout/activity_search.xml
```diff
@@ -0,0 +1,21 @@
...
+
+  <com.mapzen.pelias.widget.AutoCompleteListView
+      android:id="@+id/list_view"
+      android:layout_width="match_parent"
+      android:layout_height="match_parent"
+      android:background="@color/mz_white"
+      android:visibility="gone"
+      />
+
+</FrameLayout>
```
ecgreb: Newline

library/src/main/java/com/mapzen/android/dagger/AndroidModule.java
```diff
@@ -69,13 +69,21 @@
...
       Log.e(TAG, e.getLocalizedMessage());
     }
     return handler;
   }
+
+  /**
+   * Provide initializer for configuring {@link MapzenSearch} objects.
+   * @return
+   */
+  @Provides @Singleton public SearchInitializer provideSearchInitializer() {
```
ecgreb: Nice and clean. Perhaps we will want to introduce an initializer for tiles and routing at some point too similar to the `SearchInitializer` class.

library/build.gradle
```diff
@@ -85,6 +85,7 @@ dependencies {
   compile "com.mapzen.tangram:tangram:$tangram_version"
   compile 'com.mapzen.android:lost:1.1.1'
   compile 'com.mapzen:on-the-road:0.8.4'
+  compile('com.mapzen.android:pelias-android-sdk:0.7.2-SNAPSHOT')
```
ecgreb: I'm hoping we won't have to worry about excluding the javax `annotation-api` dependency here like we currently do in EM. I believe we ran into issues a while back with two different versions included by two different libraries.

https://github.com/mapzen/eraser-map/blob/master/app/build.gradle#L133

If indeed we don't need the exclude here we can drop the `( )` around the module descriptor.

PR #151, opened 2016-08-01
Add ability to persist MapData on orientation changes

### LINKED ISSUE #149: New client data objects after config change
MapData objects should not be persisted when the MapView is part of a view that is destroyed.

New instances should be created along with the activity.

sarahlensing:
- Adds `MapzenMap#setPersistMapData` to optionally restore state on orientation changes
- Adds `MapDataManager` and `MapStateManager` for handling restoration of map state and data
- Removes dagger package so that new state manager classes can remain package private
- Updates documentation

Closes #149


library/src/main/java/com/mapzen/android/DI.java
```diff
@@ -1,4 +1,4 @@
-package com.mapzen.android.dagger;
+package com.mapzen.android;
```
ecgreb: Why the package change?

library/src/main/java/com/mapzen/android/DataLayerType.java
```diff
@@ -0,0 +1,17 @@
+package com.mapzen.android;
+
+/**
+ * Created by sarahlensing on 7/27/16.
+ */
+public enum DataLayerType {
```
ecgreb: This list of layers looks a lot like the layer name constants declared in `OverlayManager`.

https://github.com/mapzen/android/blob/2118e5059299318e4ec4d4aa5abf1b8c967b9a53/library/src/main/java/com/mapzen/android/OverlayManager.java#L33-L42

Rather than maintain two separate lists what if we deleted the constants in `OverlayManager` and modified `DataLayerType` to make the string value be the name of the layer?

```java
/**
 * Default data layers for bundled Mapzen stylesheets including Bubble Wrap, Cinnabar, and Refill.
 * Used to add client data layers to the map and optionally persist them on configuration change.
 */
public enum DataLayerType {
  CURRENT_LOCATION("mz_current_location"),
  POLYLINE("mz_default_line"),
  POLYGON("mz_default_polygon"),
  MARKER("mz_default_point"),
  ROUTE_START_PIN("mz_route_start"),
  ROUTE_END_PIN("mz_route_destination"),
  DROPPED_PIN("mz_dropped_pin"),
  SEARCH_RESULT_PIN("mz_search_result"),
  ROUTE_PIN("mz_route_location"),
  ROUTE_LINE("mz_route_line"),
  TRANSIT_ROUTE_LINE("mz_route_line_transit"),
  TRANSIT_ROUTE_LINE_STATION_ICON("mz_route_transit_stop");

  private final String name;

  DataLayerType(String s) {
    name = s;
  }

  public String toString() {
    return name;
  }
}
```

library/src/main/java/com/mapzen/android/MapDataManager.java
```diff
@@ -0,0 +1,52 @@
+package com.mapzen.android;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Manages {@link PersistableMapData} objects for handling recreation of {@link MapData} objects
```
ecgreb: Unfortunately Javadoc links require a fully qualified class name to resolve unless that class is in the same package or included via an `import` statement.

For example `{@link PersistableMapData}` resolves since it is in the same package but `{@link MapData}` does not and instead needs to be `{@link com.mapzen.tangram.MapData}`.

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -676,4 +708,26 @@ public void drawTransitRouteLine(@NotNull List<LngLat> points, @Nullable List<Ln
   public void clearTransitRouteLine() {
     overlayManager.clearTransitRouteLine();
   }
+
+  /**
+   * By default all {@link MapData} objects are removed from the map when it is destroyed. To
+   * persist this data such as in the case of orientation changes, use this method.
+   * @param persistOnRecreation persist {@link MapData} across orientation changes
+   */
+  public void setPersistMapData(boolean persistOnRecreation) {
```
ecgreb: I wonder if we should have two separate flags that can be set: one for persisting camera state (position, zoom, rotation, tilt) and one for persisting client data sources (points, lines, polygons)? Is there any use case where you would want one and not the other?

library/src/main/java/com/mapzen/android/MapzenMap.java
```diff
@@ -676,4 +708,26 @@ public void drawTransitRouteLine(@NotNull List<LngLat> points, @Nullable List<Ln
...
+    overlayManager.setPersistMapData(persistOnRecreation);
+    mapStateManager.setPersistMapState(persistOnRecreation);
+  }
+
+  private void restoreMapState() {
```
ecgreb: Based on initial non-scientific testing (using a stopwatch timer) invoking this method seems to double the time it takes before the map is displayed on the screen from ~2 seconds to ~4 seconds which feels _really_ long.

I wonder if there is anything we can do to speed up the process of recreating the map state?

cc @blair1618 @tallytalwar

library/src/main/java/com/mapzen/android/DI.java
ecgreb: Nevermind now I see why. Its because `MapDataManager` and `MapStateManager` are now package private like we discussed ;)

Still based on the number of classes accumulating in the `com.mapzen.android` package I'm thinking we need a better way to organize things.

Perhaps we can add a package for each major feature/component. Something like this:

`com.mapzen.android.graphics` - Tangram related classes
`com.mapzen.android.search` - Pelias related classes
`com.mapzen.android.routing` - Valhalla related classes
`com.mapzen.android.location` - LOST related classes

Then all the common classes like `MapzenMap`, API key related stuff, and dependency injection could remain in `com.mapzen.android` or we could create a `com.mapzen.android.core` package.

Thoughts?

library/src/main/java/com/mapzen/android/MapzenMap.java
ecgreb: Also I'm starting to think the default here should be that the data is persisted because now the sample app feels broken when you rotate it and the map state resets each time.

library/src/main/java/com/mapzen/android/MapzenMap.java
blair1618: I would wager `setStyle(MapStyle)` is what's slow. `MapController$loadSceneFile(String)` is a heavy, _heavy_, call (this is why we have a method to do it on a background thread). We have some ideas about speeding this up for large scene files in the future, but right now it's definitely worthwhile to do whatever you can to avoid this call. Maybe we could check whether the 'persisted' style is the same as the current style?

library/src/main/java/com/mapzen/android/DI.java
sarahlensing: I like this, I think it makes sense to do and will definitely clean things up nicely for us. I would consider `MapzenMap` a `com.mapzen.android.graphics` class though, especially because it relies on some package private classes that will squarely reside in `com.mapzen.android.graphics` (OverlayManager, MapStateManager). What do you think?

library/src/main/java/com/mapzen/android/DataLayerType.java
sarahlensing: I was on the verge of doing this during the first pass but was slightly swayed not to because some `PersistableMapData` objects represent two different layer types (START_END_PIN has "mz_route_start" and "mz_route_destination". TRANSIT_ROUTE_LINE_STATION_ICON has "mz_route_line_transit" and"mz_route_transit_stop" data layers). I also wanted the switch statement where we restore data to be exhaustive but I can definitely see the convenience of maintaining only one list. I'll update.

library/src/main/java/com/mapzen/android/DI.java
ecgreb: Good point. Let's go with that. I'm thinking after this pull request is wrapped up we can handle breaking things down into packages in its own change set.

ecgreb: https://github.com/mapzen/android/issues/153

library/src/main/java/com/mapzen/android/MapDataManager.java
sarahlensing: Fixed.

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: As discussed, I'm on board with this.

Added a separate method `setPersistMapState` and have it defaulting to `true`

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: Really nice catch. I updated the code so that we only set the scene file once, in the `MapInitializer`

library/src/main/java/com/mapzen/android/MapzenMap.java
sarahlensing: Whoops, just seeing the second comment after refreshing and want to clarify- we want to persist the map state on default but not the map data, right?

Separately, we should update the sample app to persist data so that it doesnt feel broken

ecgreb:
I ran into the following crash when testing the "Route Line" demo but I think its unrelated to the recent changes. I think what happened is that I clicked the "Clear" button before `getMapAsync(...)` had completed and so `RouteLineActivity.this.map` was still null at this point.

```
Process: com.mapzen.android.sample, PID: 19977
    java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapzen.android.graphics.MapzenMap.clearRouteLine()' on a null object reference
    at com.mapzen.android.sample.RouteLineActivity$1.onClick(RouteLineActivity.java:35)
    at android.view.View.performClick(View.java:5205)
    at android.view.View$PerformClick.run(View.java:21164)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:148)
    at android.app.ActivityThread.main(ActivityThread.java:5417)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

library/src/main/java/com/mapzen/android/graphics/MapzenMap.java
```diff
@@ -179,7 +184,7 @@ public float getZoom() {
    * Set map position without animation.
    */
   public void setPosition(LngLat lngLat) {
-    mapController.setPosition(lngLat);
+    setPosition(lngLat, 0);
```
ecgreb: I'm not sure that calling `mapController.setPositionEased(lngLat, 0, EaseType.CUBIC)` with a duration of `0` is the same as calling `mapController.setPosition(lngLat)`. This could be the source of the issues we are seeing in Eraser Map with setting the position in route preview and routing mode?

blair1618: Good eye. I don't believe it's currently guaranteed that an eased call with a duration of zero will produce the exact same result as a non-eased call. We can (and probably should) make this true, but for now I'd recommend not easing with durations of zero.

tallytalwar: Nice catch. Filed an issue for us: https://github.com/tangrams/tangram-es/issues/903

PR #152, opened 2016-08-02
Dont bundle dagger compiler in apk

sarahlensing:


library/build.gradle
```diff
@@ -13,6 +13,7 @@ ext.tangram_version = "0.4"
 apply plugin: 'com.android.library'
 apply plugin: 'checkstyle'
 apply plugin: 'com.github.dcendents.android-maven'
+apply plugin: 'com.neenbedankt.android-apt'
```
ecgreb: Oh cool. This is the same plugin I have used in the past to get Dagger generated classes to resolve in IntelliJ.

I didn't realize it added another dependency scope as well.

blair1618:
Whoa cool, what is this `apt` scope?

sarahlensing:
Its like the `provided` scope but doesn't put the annotation processor on your classpath, only the generated classes!

https://bitbucket.org/hvisser/android-apt

PR #156, opened 2016-08-05
Directly call MapController methods for setting position, tilt w/out animation

sarahlensing:


blair1618:
Did this fix the issues you were seeing in eraser map? After looking more closely, tangram-es _does_ correctly handle eases of zero duration, so I'm not sure what the root of this problem would be.

sarahlensing:
Hmm thats interesting, it did fix the problem in eraser map. There must be something else going on there?

blair1618:
¯\\(ツ)/¯

PR #158, opened 2016-08-08
Adds Gradle Release plugin

### LINKED ISSUE #157: Prevent snapshot dependencies from being included in release builds
Release versions of the SDK should not depend upon snapshots of other libraries as it can introduce instability in client apps. We should update the build process to prevent this from happening in the future.

https://github.com/mapzen/android/blob/v1.0.8/library/build.gradle#L89

One approach would be to remove the snapshot repository from the SDK `build.gradle` and add it locally for development.

https://github.com/mapzen/android/blob/v1.0.8/build.gradle#L21

Alternatively we could experiment with a release plugin similar to the [Maven Release Plugin](http://maven.apache.org/maven-release/maven-release-plugin/) that will fail a release build if it includes any snapshot dependencies.

A quick search shows there are several Gradle plugins that provide similar functionality. Here is the one that appears to be most actively maintained.

https://github.com/researchgate/gradle-release

As a workaround for release v1.0.8 client apps can add the sonatype snapshot repository to their `build.gradle`.

```groovy
repsitories {
    ...
    maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
}
```

ecgreb:
* Adds release plugin to `build.gradle`.
* Sets path to `gradle.properties`.
* Adds `version` property used by gradle release plugin.
* Syncs `VERSION_NAME` property used by android-maven plugin and gradle-mvn-push.gradle.
* Triggers `uploadArchives` after successful release build.

https://github.com/researchgate/gradle-release

Closes #157

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,5 @@
+version=1.0.9-SNAPSHOT
+
 POM_ARTIFACT_ID=mapzen-android-sdk
 POM_NAME=Mapzen Android SDK
 POM_PACKAGING=aar
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -25,6 +25,7 @@ version = VERSION_NAME
 project.archivesBaseName = POM_ARTIFACT_ID
 
 release {
+  tagTemplate = 'v${version}'
   versionPropertyFile = '../gradle.properties'
   versionProperties = ['VERSION_NAME']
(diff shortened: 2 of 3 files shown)


sarahlensing:
👍 This really cleans up our release process!

PR #170, opened 2016-09-12
add metrics

### LINKED ISSUE #164: Analyze size, method count, and performance
It would be good to have the following metrics available for the Mapzen SDK:

* AAR size
* Impact on client app APK size (is this the same?)
* Method count
* Comprehensive list of dependencies including transitive dependencies (can we omit any currently in use?)
* Comprehensive list of permissions (and those that can/should be optionally added by client apps)
* Basic performance metrics for startup time, map load time, cached vs. non-cached tiles, etc.
* What else?

sarahlensing:
Adds gradle tasks and adb log output to get various metrics:

- aar size
`./gradlew aarSize`

- number of methods
`./gradlew countReleaseDexMethods`

- view dependency tree
`./gradlew library:dependencies --configuration compile`

- view required permissions
`./gradlew permissions`

- map load time
`adb logcat -s BasicMapzenActivity`

- cached / non-cached tile load times
`adb logcat -s TileHttpHandler` (must enable logging by modifying value of `DEBUG_REQUESTS` in `TileHttpHandler`)

Newly added gradle tasks are run each time master is built

TODO:
- task to analyze unused dependencies (https://github.com/mapzen/android/issues/186)

Closes #164

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -9,7 +9,7 @@ buildscript {
   dependencies {
     classpath 'com.github.dcendents:android-maven-gradle-plugin:1.3'
     classpath 'net.researchgate:gradle-release:2.4.0'
-    classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.6.0'
+    classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.6.1'
   }
 }
 
@@ -126,10 +126,9 @@ task aarSize {
 }
 aarSize.dependsOn(assembleRelease)
 
-//TODO: uncomment after merge: https://github.com/KeepSafe/dexcount-gradle-plugin/pull/133
-//dexcount {
-//  runOnEachAssemble false
-//}
+dexcount {
+  runOnEachAssemble false
+}
 
 task permissions {
   File manifest = file("src/main/AndroidManifest.xml")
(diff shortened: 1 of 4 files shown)


sample/src/main/java/com/mapzen/android/sample/BasicMapzenActivity.java
```diff
@@ -48,10 +50,15 @@
...
         savedInstanceState.getBoolean(KEY_LOCATION_ENABLED));
 
+    final float millis = System.currentTimeMillis();
+
     final MapFragment mapFragment =
         (MapFragment) getSupportFragmentManager().findFragmentById(R.id.fragment);
     mapFragment.getMapAsync(new OnMapReadyCallback() {
       @Override public void onMapReady(MapzenMap map) {
+        float now = System.currentTimeMillis();
+        float elapsedTime = now - millis;
```
ecgreb: System time should be captured in a `long` instead of a `float`. Otherwise the difference is lost in precision and `0.0` elapsed time is always reported.

Ditto for local vars `now` and `millis` above.

library/build.gradle
```diff
@@ -113,3 +115,25 @@ dependencies {
 }
 
 apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
+
+task aarSize {
+  doLast {
+    File aar = file("build/outputs/aar/mapzen-android-sdk-release.aar")
+    println ""
+    println "AAR Size: ${aar.length()} bytes"
```
ecgreb: This is great although we do still need a way to calculate the total size with transitive dependencies so when someone asks "How much bigger will the Mapzen SDK make my app?" we can give a more accurate answer.

PR #181, opened 2016-09-26
Update dependencies

ecgreb:
Updates pelias-android-sdk, on-the-road, and lost.

`OverlayManager` should wait for Lost connection callbacks before manipulating map data.

Also adds logic to make sure current `OverlayManager` (recreated when the activity is destroyed) receives the callback. Fixes crash in sample app on rotation.


msmollin:
👍  assuming the tests get passing :)

PR #204, opened 2016-10-17
Places module

### LINKED ISSUE #193: Mapzen places: new module

sarahlensing:
### Overview

Creates places module within the Android SDK project
### Proposed Changes

This PR extracts common dependency versions between the places and library modules so that they can be upgraded together. It also bundles the places module within the Mapzen Android SDK artifact that is deployed.

Closes #193


gradle.properties
```diff
@@ -1,4 +1,6 @@
 version=1.1.1-SNAPSHOT
+pelias_version=1.0.0
```
ecgreb: This is cool. Should we do pull out all the dependency version numbers into props (not just Mapzen libs)?

ecgreb:
Looking at the overall folder structure I'd like to suggest a few changes.

1. Create a new folder called `sdk` in the project root.
2. Move the current `library` and `sample` modules into the `sdk` folder.
3. Create a new folder called `places` in the project root.
4. Rename the `places` module to `library` and move it into the `places` folder.
5. Add a `sample` app module to the `places` folder which will become the places demo app.

The final directory structure would look something like the diagram below. This way each build artifact (AAR) would have its own sample app. Both would depend on the `core` module that lives off to the side in its own folder.

```
[PROJECT_ROOT]
    -> core
        -> core (module - aar)
    -> sdk
        -> library (module - aar)
        -> sample (module - apk)
    -> places
        -> library (module - aar)
        -> sample (module - apk)
```

ecgreb:
Also I am wondering if we want to take a more measured approach to moving around the existing classes.

Instead of moving every class from `library` into `core` we might think `places` will ever want... what do you think about moving each class that places will depend on in an as-needed fashion in future pull requests?

sarahlensing:
I like having an apk for each aar and agree that we could make the structure more self explanatory but I have a couple concerns with these suggestions:

1. Module names have to be unique so renaming what we have to represent this structure will result in `library` -> `sdk-library`, `sample` -> `sdk-library`, `places` -> `places-library`, `core` -> `core-library`. We wouldnt be able to have two `library` modules if we placed them in different directories.
2. Adding the extra container directory for each set of library/sample seems to break the conventional way of structuring multiple modules in Android Studio. Also, if we rename the modules to be more explanatory, it feels like the extra directory isn't necessary

What are your thoughts? Do you think renaming them to the suggestions in 1 solves the problem?

sarahlensing:
I'm on board with moving classes into the core module on an as-needed basis and will revert the changes that move them in this PR

ecgreb:
I didn't realize the module names needed to be unique even if they are in different folders. Even still I 100% agree with the suggestion to namespace them to be more explicit.

However I'm not sure we would be breaking any convention by moving our modules into subfolders of the project. For example [Robolectric has similar modules grouped together](https://github.com/robolectric/robolectric/tree/master/robolectric-shadows)  in the `shadows` subfolder and [OkHttp has all sample apps grouped together](https://github.com/square/okhttp/tree/master/samples) in the `samples` folder.

I think we should organize them in whichever way makes the most sense for our project.

### So based on these examples the question becomes which of the following do we want?

#### 1. All modules at the root level?
```
[PROJECT_ROOT]
    -> core (aar)
    -> mapzen-android-sdk (aar)
    -> mapzen-android-sdk-sample (apk)
    -> mapzen-places-api (aar)
    -> mapzen-places-api-sample (apk)
```

#### 2. All modules in subfolders?
 ```
[PROJECT_ROOT]
    -> core
        -> core (aar)
    -> mapzen-android-sdk
        -> mapzen-android-sdk (aar)
        -> mapzen-android-sdk-sample (apk)
    -> mapzen-places-api
        -> mapzen-places-api (aar)
        -> mapzen-places-api-sample (apk)
```

#### 3. Or some combination of both?
_For example grouping all sample projects._
```
[PROJECT_ROOT]
    -> core (aar)
    -> mapzen-android-sdk (aar)
    -> mapzen-places-api (aar)
    -> samples
        -> mapzen-android-sdk-sample (apk)
        -> mapzen-places-api-sample (apk)
```

**Comment by ecgreb** [comment]
Also probably worth noting we would never actually "publish" the `core` aar as its own library. Its for internal organization of the code in this project only.

**Comment by sarahlensing** [comment]
I'd vote for the options in this order: 
3. Or some combination of both?
1. All modules at the root level?
2. All modules in subfolders?

I like keeping all the library modules at the same (root) level because it emphasizes their connectedness to each other & their first class status as libraries. The samples are secondary and feel better grouped in a separate folder (or at the root for now as we only have 2 of them).

What do others think? cc/ @msmollin @blair1618 @tallytalwar @karimnaaji

**Comment by ecgreb** [comment]
@sarahlensing I like your suggestion and rationale for how to organize the modules. All libs at root level. Samples in a subfolder.

I'm also interested to hear what others think about best way to lay out the project so its easiest to maintain and also how to build and understand for contributors in the future.

**Comment by ecgreb** [comment]
If there are no objections let's move ahead with the changes discussed above:

1. Renaming modules to be more explicit
2. Keeping all library modules at root level
3. Creating a new subfolder for sample apps
4. Keeping all classes in the SDK module and only moving them into core module once they are required by classes in the places module

Sound good?

**Comment by ecgreb** [comment]
New module layout looks good. I am able to build and install both AARs into my local Maven repo and both sample apps when testing locally 👍 

The only issue I ran into is when trying to consume the `mapzen-android-sdk` AAR in a local build of Eraser Map. The build failed on the following error:

```
Tue Dec 06 11:07 AM #2066 /usr/local/code/eraser-map [master] 
$ ./gradlew clean installDevDebug --refresh-dependencies
Download https://mint.splunk.com/gradle/com/splunk/mint/mint/4.2.1/mint-4.2.1.pom
File /home/ecgreb/.android/repositories.cfg could not be loaded.

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_prodDebugApkCopy'.
   > Could not find mapzen-android-sdk:core:unspecified.
     Required by:
         eraser-map:app:0.13.0-SNAPSHOT > com.mapzen:mapzen-android-sdk:1.2.0-SNAPSHOT

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 1 mins 17.531 secs
```

PR #219, opened 2016-11-02
Disable current location layer should remove persisted map data

ecgreb:
### Overview

Fixes bug where more than one instance of the current location layer would be persisted leading to inconsistent results when map data state is restored.

### Proposed Changes

When current location layer is disabled all instances of `DataLayerType.CURRENT_LOCATION` are removed from `MapDataManager`. This prevents the situation where more than one current location layer is persisted, some with `PersistableMapData.locationEnabled` set to `true` and others `false`.

This change set also reduces the scope of `PersistableMapData` constructors to package local since this class is only intended to be used internally and adds javadoc descriptions for constructor params.

Closes https://github.com/mapzen/eraser-map/issues/753

### DIFF
--- a/library/src/main/java/com/mapzen/android/graphics/OverlayManager.java
+++ b/library/src/main/java/com/mapzen/android/graphics/OverlayManager.java
@@ -162,20 +162,26 @@ public class OverlayManager implements TouchInput.PanResponder {
    */
   public void setMyLocationEnabled(boolean enabled, boolean persistMapData) {
     myLocationEnabled = enabled;
+    handleTangramMapDataChanges(enabled);
+    updateCurrentLocationPersistableMapData(enabled, persistMapData);
+    handleMyLocationEnabledChanged();
+  }
 
+  private void handleTangramMapDataChanges(boolean enabled) {
     if (enabled) {
       addCurrentLocationMapData();
-      if (persistMapData) {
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/mapzen/android/graphics/OverlayManager.java
```diff
@@ -161,14 +161,18 @@ public void setMyLocationEnabled(boolean enabled) {
    * Optionally persists the map data to the {@link MapDataManager}.
    */
   public void setMyLocationEnabled(boolean enabled, boolean persistMapData) {
```
sarahlensing: We should also add a check to see if the value of `enabled` changes so that we avoid adding two `PersistableMapData` objects when `enabled` is `true`. 

```
boolean changed = myLocationEnabled != enabled;
    if (!changed) {
      return;
    }
    myLocationEnabled = enabled;
```

ecgreb: I solved for this case by always removing the current location persisted map data when `persistMapData` is `true`. Then I only re-add that layer if `enabled` is also `true`.

Also took this chance to break persistence logic and Tangram map data UI logic into separate private methods.

sarahlensing: Awesome

PR #226, opened 2016-11-29
Label pick listener

sarahlensing:
### Overview
Adds `LabelPickListener` interface for label selection.

### Proposed Changes
This PR upgrades tangram and updates the internal tap responder to call `MapController#pickLabel` if a `LabelPickListener` has been set on the map.


sample/src/main/java/com/mapzen/android/sample/LabelPickListenerActivity.java
```diff
@@ -0,0 +1,62 @@
...
+
+    LabelPickListener listener = new LabelPickListener() {
+        @Override
+        public void onLabelPicked(LabelPickResult result, float positionX, float positionY) {
+            if (result == null) {
+                return;
+            }
+            double longitude = result.getCoordinates().longitude;
+            double latitude = result.getCoordinates().latitude;
+            map.addMarker(new Marker(longitude, latitude));
```
ecgreb: We should probably clear previous markers when a new label is selected.

sample/src/main/java/com/mapzen/android/sample/LabelPickListenerActivity.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+    };
+
+    FeaturePickListener featurePickListener = new FeaturePickListener() {
```
ecgreb: Seems like FeaturePickListener should be removed here since there is a separate `FeaturePickListenerActivity` to demo this behavior?

sample/src/main/java/com/mapzen/android/sample/LabelPickListenerActivity.java
ecgreb: Also it would be nice to toast the name of the label selected if this is available in the properties?

PR #232, opened 2016-12-14
Initial PlacePicker API

sarahlensing:
### Overview
Adds basic flow for `PlacePicker` api. Client applications can use the intent created by the api to launch a map. When the user selects a POI, a dialog will display the POI's title and allow the user to confirm their selection. Upon confirmation, the activity will set a result and finish.

### Proposed Changes
There will be a follow up PR which expands the functionality of this work. It will:
- Retrieve more details about the place from `Pelias`
- Construct an `Intent` with the place's information when the user confirms selection
- Implement `PlacePicker` static methods to retrieve information about the place
- Add a search bar and invoke the autocomplete activity when pressed
- Center the map on the user's current location or the pre-determined bounds

Further work will do the following:
- Add a list of nearby places given the map's current position
- Add a centered pin overlay to the map
- Dynamically load more POIs as the user pans the map


core/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.mapzen.core">
+    package="com.mapzen">
```
ecgreb: Why the package name change for core module?

core/src/main/java/com/mapzen/android/graphics/MapView.java
```diff
@@ -49,8 +49,8 @@ public MapView(Context context, AttributeSet attrs) {
   }
 
   private void initDI(Context context) {
-    DI.init(context);
-    DI.component().inject(this);
+    CoreDI.init(context);
```
ecgreb: We should replace `context` with `context.getApplicationContext()` when initializing dependency injection to prevent a memory leak. The `Context` associated with the `MapView` is tied to the `Activity` that created it.

I know we have been using the `MapView` context for this purpose for a while now but this just hit me when re-reading it.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlacePickerPresenter.java
```diff
@@ -0,0 +1,27 @@
...
+   * Called when a POI on the map has been selected by the user.
+   * @param properties
+   */
+  void onLabelPicked(Map<String, String> properties);
+
+  /**
+   * Called when a user confirms their selection in the alert dialog that is displayed after
+   * a POI has been picked.
+   */
+  void onPlaceSelected();
```
ecgreb: The name `onPlaceConfirmed()` seems a little more explicit about when this gets called in the flow?

core/src/main/AndroidManifest.xml
sarahlensing: I thought it might make more sense to have a generic package name as we move public facing classes like `MapzenMap` into `core`. I wanted the existence of the core module to be as transparent as possible. What are your thoughts on this?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlacePickerPresenter.java
sarahlensing: Agreed, thats a better name...I'll update it.

core/src/main/AndroidManifest.xml
ecgreb: Either one is fine with me. I was just curious if you ran into any issues the old package.

PR #239, opened 2017-01-07
Add Compass button into MapView

rwrx:
### Overview
Added Compass button into MapView with two simple tests. Compass button was taken from Eraser Map. Behavior of button is like in Google Maps API for Android - it will be shown when map rotation is not equal to 0 and when this button is clicked map rotation will change to 0 and button will automatically hide.


core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
```diff
@@ -814,4 +871,13 @@ private MapData addPointToMarkerMapData(Marker marker) {
     findMe.setActivated(false);
     return false;
   }
+
+  @Override public boolean onRotate(float x, float y, float rotation) {
+    final View compass = mapView.findViewById(R.id.mz_compass);
```
sarahlensing: Looks like we can use the `mapView.getCompass()` api here

core/src/main/res/layout/mz_view_map.xml
```diff
@@ -11,6 +11,20 @@
       android:layout_height="match_parent"
       />
 
+  <com.mapzen.android.graphics.CompassView
+      android:id="@+id/mz_compass"
+      android:layout_width="@dimen/mz_compass_button_width"
+      android:layout_height="@dimen/mz_compass_button_height"
+      android:layout_alignParentLeft="true"
```
sarahlensing: We should align the compass to be parent right by default since this is how both GMaps and EraserMap work

core/src/test/java/com/mapzen/android/graphics/MapzenMapTest.java
```diff
@@ -293,10 +293,10 @@
     verify(overlayManager).onFling(1f, 2f, 3f, 4f);
   }
 
-  @Test public void setRotateResponder_shouldInvokeMapController() {
+  @Test public void setRotateResponder_shouldNotInvokeMapController() {
```
sarahlensing: This is good but lets also add 2 more tests to ensure that the `OverlayManager` and any external `RotateResponder` objects are invoked:

```
@Test public void setRotateResponder_shouldInvokeOverlayManager() {
    TestRotateResponder rotateResponder = new TestRotateResponder();
    map.setRotateResponder(rotateResponder);
    map.internalRotateResponder.onRotate(1, 2, 0.5f);
    verify(overlayManager).onRotate(1, 2, 0.5f);
  }

  @Test public void setRotateResponder_shouldInvokeRotateResponder() {
    TestRotateResponder rotateResponder = new TestRotateResponder();
    map.setRotateResponder(rotateResponder);
    map.internalRotateResponder.onRotate(1, 2, 0.5f);
    assertThat(rotateResponder.rotated).isTrue();
  }
```

And then update `TestRotateResponder` to look like this:
```
public class TestRotateResponder implements TouchInput.RotateResponder {
  boolean rotated = false;

  @Override public boolean onRotate(float x, float y, float rotation) {
    rotated = true;
    return false;
  }
}
```

core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
rwrx: Sure, I did not realized this. I have basically looked at how are onFling() and onPan() methods written. So in these methods it will be better to replace mapView.findViewById(R.id.mz_find_me) with getFindMe() too.

core/src/main/res/layout/mz_view_map.xml
rwrx: Ah yes, I didn't know that in GMaps it is on the right. But in Google Maps API it is on left side. I put it to the right because I needed to retain consistency with Google Maps API and not to confuse users. Also I have added zoom buttons like in Google Maps API and moved location button to the top right so this is why I put compass on left top. But yes, these are only my notes why I did this.

core/src/test/java/com/mapzen/android/graphics/MapzenMapTest.java
rwrx: Great! Is there something that I need to do? Or you will incorporate all these changes?

core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
sarahlensing: Agreed, lets also change those methods

core/src/test/java/com/mapzen/android/graphics/MapzenMapTest.java
sarahlensing: Can you add these two tests into `MapzenMapTest`, update `TestRotateResponder`, and then run `./gradlew testDebug` to make sure that the tests pass? Thank you!

core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
rwrx: Ok, I tried to replace that in onFling() and onPan() with mapView.getFindMe(), but two tests onFling_shouldDeactivateFindMe() and onPan_shouldDeactivateFindMe() are failing because there is a null pointer exception on findeMe.setActivated(false);. I don't know how mocking works but this seems to be a problem that MapView stored in OverlayManager has field findMe not initialized.

ecgreb:
Hmm looks like Circle CI is not running builds for fork pull requests on the SDK repo now either I will fiddle with settings and see if I can get them working again.

core/src/main/res/layout/mz_view_map.xml
sarahlensing: I didn't realize the sdk differed from the app :) Given that, I retract my comment, the compass should be in the top left. I've created a ticket here to provide an alternate configuration for developers who want consistency with the app:
https://github.com/mapzen/android/issues/241

core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
sarahlensing: I created an issue for this and we can address it in a separate PR: https://github.com/mapzen/android/issues/242

core/src/main/res/layout/mz_view_map.xml
ecgreb: +1 for providing both overlay configuration options.

I don't mind merging this implementation as-is then re-introducing the "classic" overlay layout option for Eraser Map and other apps that would prefer the original button placement.

core/src/test/java/com/mapzen/android/graphics/MapzenMapTest.java
ecgreb: I agree once these tests are added we are good to merge. Tests should now automatically run on Circle CI when the next commit is made.

rwrx:
I have incorporated changes according to review. Circle CI is now finally passing.

Issue #196, opened 2016-10-12
Mapzen places: autocomplete UI


ecgreb:

ecgreb:
First part implemented by #235. Outstanding items below.

- [ ] Finish `PlaceAutocompleteActivity` once an item has been selected and populate search box text.
- [ ] Provide back arrow in `PlaceAutocompleteActivity` toolbar to exit without making a selection.
- [ ] Clicking soft keyboard search button should dismiss keyboard (currently crashes).
- [ ] Add `MODE_OVERLAY` for `PlaceAutocompleteActivity` giving the option to be styled as a dialog.
- [ ] Add `PlaceSelectionListener` to deliver selection results.
- [ ] Add ability to launch `PlaceAutocompleteActivity` using `PlaceAutocomplete.IntentBuilder`.
- [ ] Add ability to set bounds bias to focus results on a given bounding box.
- [ ] Add ability to filter by type (address, POI, etc.)
- [ ] Add autocomplete overlay to `PlacePicker` component.

PR #245, opened 2017-01-12
Places Autocomplete (part 2)

sarahlensing:
### Overview
This adds more functionality to the `PlaceAutocomplete` API. Developers can set a bounding box to limit autocomplete results and use the static methods in `PlaceAutocomplete` to launch the autocomplete activity and retrieve place results. This also integrates the autocomplete UI into the `PlacePicker` UI.

### Proposed Changes
The autocomplete search view has been refactored into `PlaceAutocompleteView`. `AutocompleteFilter` is now `Parcelable` so that it can be passed as an intent extra. `PlaceAutocompleteActivity` respects any `BoundingBox` set. The autocomplete fragment and presenter are now passing around a proper `Place` object and there is a `PlaceSelectionListener` to handle notifying relevant objects when a `Place` has been selected from the autocomplete fragment.

There will be a follow up PR with more changes as per: https://github.com/mapzen/android/issues/196

### DIFF
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
@@ -11,7 +11,9 @@ import android.content.Context;
 import android.content.Intent;
 
 import static com.mapzen.places.api.internal.PlaceIntentConsts.EXTRA_BOUNDS;
+import static com.mapzen.places.api.internal.PlaceIntentConsts.EXTRA_FILTER;
 import static com.mapzen.places.api.internal.PlaceIntentConsts.EXTRA_PLACE;
+import static com.mapzen.places.api.internal.PlaceIntentConsts.EXTRA_STATUS;
 
 /**
  * Main entry point for the autocomplete API. A search widget is presented and users can select
@@ -21,9 +23,6 @@ public class PlaceAutocomplete {
 
   public static final int MODE_FULLSCREEN = 1;
   public static final int MODE_OVERLAY = 2;
-
-  public static final String EXTRA_STATUS = "extra_status";
-  public static final String EXTRA_FILTER = "extra_filter";
 
   /**
    * Retrieves a {@link Place} object from the {@link Intent}.
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
@@ -26,7 +26,7 @@ public class PlaceAutocomplete {
   public static final String EXTRA_FILTER = "extra_filter";
 
   /**
-   * Returns a {@link Place} object.
+   * Retrieves a {@link Place} object from the {@link Intent}.
    * @param context
    * @param intent
    * @return
@@ -38,6 +38,12 @@ public class PlaceAutocomplete {
     return intent.getParcelableExtra(EXTRA_PLACE);
   }
 
+  /**
+   * Retrieves a {@link Status} object from the {@link Intent}.
+   * @param context
+   * @param intent
+   * @return
+   */
   public static Status getStatus(Context context, Intent intent) {
     if (intent == null) {
       return null;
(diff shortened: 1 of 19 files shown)


mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
```diff
@@ -0,0 +1,105 @@
...
+ * Main entry point for the autocomplete API. A search widget is presented and users can select
+ * autocomplete results. Unlike the {@link PlacePicker} API, a map is not shown.
+ */
+public class PlaceAutocomplete {
+
+  public static final int MODE_FULLSCREEN = 1;
+  public static final int MODE_OVERLAY = 2;
+
+  public static final String EXTRA_STATUS = "extra_status";
+  public static final String EXTRA_FILTER = "extra_filter";
```
ecgreb: Would these be better in `PlaceIntentConsts` to avoid adding them to the public API?

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocompleteView.java
```diff
@@ -0,0 +1,99 @@
...
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+/**
+ * Compound view which displays search icon, autocomplete text view and clear button.
+ */
+public class PlaceAutocompleteView extends LinearLayout {
```
ecgreb: This class should not be part of the public API. Can it be moved to internal package or made package private scope?

Or as an alternative do we want to consider moving this class into `pelias-android-sdk` if we are going to be reworking the search view component that resides there? Could have one implementation that works in both places?

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceSelectionListener.java
```diff
@@ -0,0 +1,23 @@
...
+ * Interface for receiving information about a {@link Place} after it is selected from a
+ * {@link PlaceAutocompleteFragment}.
+ */
+public interface PlaceSelectionListener {
+
+  /**
+   * Called when the user selects a {@link Place}.
+   * @param place
+   */
+  abstract void onPlaceSelected(Place place);
```
ecgreb: No need for `abstract` in interface method declaration (implied).

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceSelectionListener.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Called when an error occurs while trying to select a {@link Place}.
+   * @param status
+   */
+  abstract void onError(Status status);
```
ecgreb: No need for `abstract` in interface method declaration (implied).

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocomplete.java
sarahlensing: Yes, I think so. Good call. Moved.

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocompleteView.java
sarahlensing: It can be moved to the internal package, we still need the public scope because it's accessed by the `PlaceAutocompleteFragment`.

I think we should keep the class in this module because the `pelias-android-sdk` will not know about the Place activities we launch. After that reworking is done, we could update this class to inherit from the `PeliasSearchView`, remove the redundant assets from this module, and override what happens when the user clicks the input text.

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceSelectionListener.java
sarahlensing: Rm'd

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceSelectionListener.java
sarahlensing: Rm'd

mapzen-places-api/src/main/java/com/mapzen/places/api/ui/PlaceAutocompleteView.java
ecgreb: 👍

PR #246, opened 2017-01-17
Autocomplete pt3

sarahlensing:
### Overview
This PR adds support for fetching details about a `Place` when using the Autocomplete API. It also allows setting a `BoundingBox` when launching this UI. The LOST library has been updated to make `Status` objects `Parcelable`, this PR now passes them back in the `Intent` data. Lastly, this adds support for filtering results using the `AutocompleteFilter` object.

### Proposed Changes
'OnPlaceDetailsFetchedListener' interface has been updated to pass a `Place` object back. `AutocompleteDetailFetchListener` has been added to handle notifying the controller when place details are successfully fetched (or fail to be fetched). `AutocompleteFilter` has been built out and and made functional. `PeliasFilterMapper` handles mapping `AutocompleteFilter` filters to parameters accepted by `Pelias`.

### DIFF
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
@@ -11,14 +11,6 @@ import java.util.Locale;
  */
 public interface Place extends Parcelable {
 
-  public static final int TYPE_LOCALITY = 1009;
-  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_3 = 1003;
-  public static final int TYPE_SUBLOCALITY = 1022;
-  public static final int TYPE_POSTAL_CODE = 1015;
-  public static final int TYPE_COUNTRY = 1005;
-  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_1 = 1001;
-  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_2 = 1002;
-
   /**
    * Returns human readable address for this place.
    * @return
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
@@ -11,6 +11,14 @@ import java.util.Locale;
  */
 public interface Place extends Parcelable {
 
+  public static final int TYPE_LOCALITY = 1009;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_3 = 1003;
+  public static final int TYPE_SUBLOCALITY = 1022;
+  public static final int TYPE_POSTAL_CODE = 1015;
+  public static final int TYPE_COUNTRY = 1005;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_1 = 1001;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_2 = 1002;
+
   /**
    * Returns human readable address for this place.
    * @return
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
@@ -1,6 +1,7 @@
 package com.mapzen.places.api;
 
 import android.net.Uri;
+import android.os.Parcelable;
 
 import java.util.List;
 import java.util.Locale;
@@ -8,7 +9,7 @@ import java.util.Locale;
 /**
  * Represents a physical place selected from a {@link com.mapzen.android.graphics.MapzenMap}.
  */
-public interface Place {
+public interface Place extends Parcelable {
 
   /**
    * Returns human readable address for this place.
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/AutocompleteDetailFetchListener.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/AutocompleteDetailFetchListener.java
@@ -1,8 +1,6 @@
 package com.mapzen.places.api.internal;
 
 import com.mapzen.android.lost.api.Status;
-import com.mapzen.android.lost.internal.DialogDisplayer;
-import com.mapzen.android.lost.internal.SettingsDialogDisplayer;
 import com.mapzen.places.api.Place;
 
 /**
@@ -12,7 +10,6 @@ import com.mapzen.places.api.Place;
 class AutocompleteDetailFetchListener implements OnPlaceDetailsFetchedListener {
 
   final PlaceAutocompleteController controller;
-  final DialogDisplayer dialogDisplayer = new SettingsDialogDisplayer();
 
   /**
    * Construct new {@link AutocompleteDetailFetchListener} and set its controller.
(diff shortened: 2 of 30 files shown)


samples/mapzen-places-api-sample/src/main/res/layout/activity_autocomplete.xml
```diff
@@ -1,18 +1,31 @@
...
+    android:orientation="vertical"
     android:paddingBottom="@dimen/activity_vertical_margin"
     android:paddingLeft="@dimen/activity_horizontal_margin"
     android:paddingRight="@dimen/activity_horizontal_margin"
     android:paddingTop="@dimen/activity_vertical_margin"
     >
 
-  <fragment
-      android:id="@+id/place_autocomplete_fragment"
-      android:name="com.mapzen.places.api.ui.PlaceAutocompleteFragment"
```
ecgreb: I think we should keep the `PlaceAutocompleteFragment` in this example since it shows off the reusable search bar UI component.

I'm guessing the reason you swapped the fragment for a custom button is to show using the `IntentBuilder` with an `AutocompleteFilter` but I think we can kill two birds with one stone here.

The `PlaceAutocompleteFragment` API should be updated to also allow setting an `AutocompleteFilter` (and `LatLngBounds`). Then internally the fragment can make use of `PlaceAutocomplete.IntentBuilder` to launch the `PlaceAutocompleteActivity` on tap.

This way we can test both implementations with a single demo activity.

What do you think?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/AutocompletePendingResult.java
```diff
@@ -32,6 +34,7 @@
   private final String query;
   private final LatLngBounds bounds;
   private final AutocompleteFilter filter;
+  private final DialogDisplayer dialogDisplayer = new SettingsDialogDisplayer();
```
ecgreb: It seems like an unfortunate side effect of the Lost `SettingsApi` implementation that the `Places` API would need to know about a `DialogDisplayer`. We should rethink how to better encapsulate this behavior moving forward.

In the meantime do we need to return a real `SettingsDialogDisplayer` in this class or can we just make this param `null` when creating a new `Status`?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/AutocompleteDetailFetchListener.java
```diff
@@ -0,0 +1,39 @@
...
+import com.mapzen.places.api.Place;
+
+/**
+ * Handles notifying the {@link PlaceAutocompleteController} after a place's details have either
+ * been successfully retrieved or failed to be retrieved.
+ */
+class AutocompleteDetailFetchListener implements OnPlaceDetailsFetchedListener {
+
+  final PlaceAutocompleteController controller;
+  final DialogDisplayer dialogDisplayer = new SettingsDialogDisplayer();
```
ecgreb: Same comment as `AutocompletePendingResult` can we just use `null` instead of a real `DialogDisplayer` until we come up with a better way to scope this class to the Lost `SettingsApi`?

mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
```diff
@@ -1,14 +1,23 @@
...
-public interface Place {
+public interface Place extends Parcelable {
+
+  public static final int TYPE_LOCALITY = 1009;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_3 = 1003;
+  public static final int TYPE_SUBLOCALITY = 1022;
+  public static final int TYPE_POSTAL_CODE = 1015;
+  public static final int TYPE_COUNTRY = 1005;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_1 = 1001;
+  public static final int TYPE_ADMINISTRATIVE_AREA_LEVEL_2 = 1002;
```
ecgreb: Are we able to effectively determine place type at this time when fetching details from Pelias?

Or should we omit these constant values in v1 so as to not mislead developers into thinking this data will be valid in the response?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceAutocompleteController.java
```diff
@@ -27,4 +28,10 @@
    * @return
    */
   LatLngBounds getBounds();
+
+  /**
+   * Return the filter that should be used to limit autocomplete results.
+   * @return
+   */
+  AutocompleteFilter getAutocompleteFilter();
```
ecgreb: This is kind of semantic point, but typically in MVP the Presenter does not "get" anything from the View/Controller. But rather the View/Controller pushes data to the Presenter when it is received as input then acts as a listener waiting for a callback with further instructions.

Rather than having two methods `getBounds()` and `getAutocompleteFilter()` here a more idiomatic way to achieve the same result would be to parse the incoming `Intent` extras in `PlaceAutocompleteActivity.onCreate()` and set the `LatLngBounds` and `AutocompleteFilter` on the `PlaceAutocompletePresenter` as soon as they are available to be stored in a field until needed.

This also has the added benefit of making the Presenter easier to test as you can set these values from the test method rather than registering a stub View/Controller to supply the values on demand.

samples/mapzen-places-api-sample/src/main/java/com/mapzen/places/api/sample/AutocompleteDemoActivity.java
```diff
@@ -1,12 +1,65 @@
...
 import android.support.annotation.Nullable;
 import android.support.v7.app.AppCompatActivity;
+import android.view.View;
+import android.widget.Button;
+import android.widget.TextView;
+
+import static com.mapzen.places.api.AutocompleteFilter.TYPE_FILTER_ADDRESS;
+import static com.mapzen.places.api.AutocompleteFilter.TYPE_FILTER_ESTABLISHMENT;
+import static com.mapzen.places.api.AutocompleteFilter.TYPE_FILTER_GEOCODE;
+import static com.mapzen.places.api.AutocompleteFilter.TYPE_FILTER_REGIONS;
```
sarahlensing: Eek! Added the checkstyle task and removed unused imports.

samples/mapzen-places-api-sample/src/main/res/layout/activity_autocomplete.xml
sarahlensing: You read my mind...I was also trying to show where to handle callbacks. Updated `PlaceAutocompleteFragment`, restored the fragment to the demo, and added stubs for the callbacks (though with this example they are slightly different than before aka an explicit listener is set instead of callbacks received in `onActivityResult`)

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/AutocompletePendingResult.java
sarahlensing: Rm'd and replaced with `Status(int statusCode)` constructor

mapzen-places-api/src/main/java/com/mapzen/places/api/Place.java
sarahlensing: Created a ticket (https://github.com/mapzen/android/issues/251) for this and removed these values for the time being.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceAutocompleteController.java
sarahlensing: Refactored. That change definitely did clean up testing the Presenter.

PR #258, opened 2017-01-18
Compass button

rwrx:
Fixed problem with package private constructor for CompassView which caused crashing of Mapzen Android SDK.

### DIFF
--- a/core/src/main/java/com/mapzen/android/graphics/CompassView.java
+++ b/core/src/main/java/com/mapzen/android/graphics/CompassView.java
@@ -13,6 +13,7 @@ import com.mapzen.R;
  */
 public class CompassView extends RelativeLayout {
 
+<<<<<<< HEAD
   /**
    * Durations in milliseconds for rotation and fade out animations.
    */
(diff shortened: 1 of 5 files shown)


ecgreb:
Thanks @rwrx ! The failure on Circle CI is not your fault there is a compile issue with the core module on master right now.

Still you should be able to simplify the commit history in this pull request by running the following commands to rebase against master.

```
git remote add mapzen git@github.com:mapzen/android.git
git fetch mapzen
git rebase -i mapzen/master
```

You will need to force push your changes to GitHub since you are rewriting the history of this branch.

```
git push origin -f compass-button
```

rwrx:
Hope it is ok now.

ecgreb:
Unfortunately it looks like there were some conflicts that needed to be resolved as indicated by the following tokens in the diff.

```
+<<<<<<< HEAD
 +=======
```

At this point it might be cleaner to just create a new branch based on mapzen/master that has only the scope change to the `CompassView` constructor and create a new pull request.

rwrx:
Ok, I have created new pull request. I hope that finally it will be ok.

Also why is in mapzen-android-sdk build.gradle file as dependency `compile 'com.mapzen:mapzen-core:0.0.1-SNAPSHOT'`? Because always when I want to test some changes to core module I must replace it with `compile project(':core')`.

ecgreb:
Good question! When we originally setup the core module we did it with `compile project(':core')` but for some reason transitive dependencies from `core` were not getting included in the `sdk` and `places` artifacts when we deployed the AAR and they would fail at runtime.

Depending on an external AAR for `core` solved the transitive dependency issue but unfortunately created other workflow issues like the one you described (also I would rather not have to deploy the `core` artifact to maven central).

I created a new pull request to switch back to using `compile project(':core')` for the `core` module dependency. I tested building the SDK snapshot AAR, deploying it to my local maven repo, and including it in Eraser Map and all seems to be working fine so let's hope the issue with transitive dependencies has now been resolved.

https://github.com/mapzen/android/pull/259

ecgreb:
Closing in favor of https://github.com/mapzen/android/pull/261

PR #259, opened 2017-01-19
Project dependencies

ecgreb:
### Overview

Replace core module external AAR dependency with local project dependency

### Proposed Changes

Also upgrades Android Gradle plugin to version 2.2.3

PR #261, opened 2017-01-19
Compass button fix

rwrx:
CompassView constructor change to public.


ecgreb:
This pull request still includes several `*.orig` files from conflicts in the rebase. These should be removed.

Also the history still includes a number of commits that could be removed with a clean rebase against master but this is less of an issue as they can be squashed into a single commit when the pull request is merged.

rwrx:
Ok, but I see that you have added that "public" to CompassView constructor in #259 so this pull request could be closed in favor of that pull request.

ecgreb:
Yes that that is true I did that so I could test the sample apps. Feel free to close this one if you would prefer.

rwrx:
Ok. Closing.

PR #264, opened 2017-01-20
Build tools 25.0.1

sarahlensing:
Upgrade build tools to 25.0.1


rwrx:
If I may say, build tools are already available in 25.0.2 version.

sarahlensing:
@rwrx Thanks for the heads up, I've updated to 25.0.2

PR #276, opened 2017-01-26
Support Place placeType

### LINKED ISSUE #251: Implement Place Type
Mapzen `Place` objects should have a `placeType` value depending on the Pelias layer they belong to (https://mapzen.com/documentation/search/search/#filter-by-data-type). 

List of possible `placeType` values: https://developers.google.com/android/reference/com/google/android/gms/location/places/Place

For some Pelias layers (`locality`, `country`, `address`), these values map cleanly to existing type values. Others will (`macrocounty`, `macroregion`) require some testing.

sarahlensing:
### Overview
This adds support for mapping `Pelias` layer strings returned from the server to `Place#placeType` ints expected by developers. It also refactors (and adds tests for) some of the `Place` fetch-related classes. [place-type-tests.txt](https://github.com/mapzen/android/files/733576/place-type-tests.txt) lists the `Pelias` http request I made, the name of the corresponding `GooglePlace` I selected, and the values I observed that place object to have. 

### Proposed Changes
- Adds `PeliasLayerConsts` to represent all possible layers that the server can return
- Refactors `PeliasFilterMapper`to use these layer values
- Add supported place types to `Place` interface
- Adds `PeliasLayerToPlaceTypeConverter` to handle converting layers returned by server into recognized `Place` type values. Adds test coverage for this class
- Extracts handling of successful/failed `Pelias` requests out of `PeliasPlaceDetailFetcher` and into new `PeliasCallbackHandler` class. Adds test coverage for this class
- Extracts conversion of `Pelias` `Feature` objects into `Place` objects out of `PeliasPlaceDetailFetcher` and into new `PeliasFeatureToPlaceConverter` class. Adds test coverage for this class
- Adds test coverage for `PeliasPlaceDetailFetcher`

Closes #251

### DIFF
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
@@ -1,0 +1,60 @@
+package com.mapzen.places.api.internal;
+
+import com.mapzen.pelias.gson.Feature;
+import com.mapzen.places.api.LatLng;
+import com.mapzen.places.api.LatLngBounds;
+import com.mapzen.places.api.Place;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Handles converting Pelias {@link Feature} objects into {@link Place} objects for
+ * {@link PeliasCallbackHandler}.
+ */
+class PeliasFeatureToPlaceConverter {
+
+  private PeliasLayerToPlaceTypeConverter layerConverter;
+
+  /**
+   * Constructor.
+   */
+  PeliasFeatureToPlaceConverter() {
+    layerConverter = new PeliasLayerToPlaceTypeConverter();
+  }
+
+  /**
+   * Construct a {@link Place} object from the given feature.
+   * @param feature
+   * @return
+   */
+  Place getFetchedPlace(Feature feature) {
+    //TODO: fill in missing values
+    final CharSequence address = feature.properties.label;
+    final String id = feature.properties.id;
+    final LatLng latLng = new LatLng(feature.geometry.coordinates.get(1),
+        feature.geometry.coordinates.get(0));
+    final Locale locale = Locale.US;
+    final CharSequence name = feature.properties.name;
+    final List<Integer> placeTypes = new ArrayList<>();
+    if (feature.properties.layer != null) {
+      placeTypes.addAll(layerConverter.getPlaceTypes(feature.properties.layer));
+    }
+    final LatLngBounds viewport = new LatLngBounds(latLng, latLng);
+    return new PlaceImpl.Builder()
+        .setAddress(address)
+        .setAttributions(null)
+        .setId(id)
+        .setLatLng(latLng)
+        .setLocale(locale)
+        .setName(name)
+        .setPhoneNumber(null)
+        .setPlaceTypes(placeTypes)
+        .setPriceLevel(0)
+        .setRating(0)
+        .setViewPort(viewport)
+        .setWebsiteUri(null)
+        .build();
+  }
+}
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
@@ -16,12 +16,14 @@ import java.util.Locale;
 class PeliasFeatureToPlaceConverter {
 
   private PeliasLayerToPlaceTypeConverter layerConverter;
+  private PointToBoundsConverter pointConverter;
 
   /**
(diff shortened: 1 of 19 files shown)


mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
```diff
@@ -0,0 +1,60 @@
...
+    layerConverter = new PeliasLayerToPlaceTypeConverter();
+  }
+
+  /**
+   * Construct a {@link Place} object from the given feature.
+   * @param feature
+   * @return
+   */
+  Place getFetchedPlace(Feature feature) {
+    //TODO: fill in missing values
```
ecgreb: We need some way to communicate to developers which fields in the `Place` object are currently supported and which are not. Perhaps some combination of the following:

1. Documentation - general documentation in README, /docs folder
2. Javadoc - explicitly call out in Javadoc which methods are not yet supported ex. `Place#getAttributions()`
3. Exception - throw a Runtime exception with message "Sorry, not yet supported" for unsupported fields
4. Return value - have a specific error code returned for fields not yet supported. ex. Strings could be "Not yet supported" and numbers could be `-1`?
5. Comment out unsupported methods - if they don't exist they cannot be called.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
```diff
@@ -0,0 +1,60 @@
...
+    final String id = feature.properties.id;
+    final LatLng latLng = new LatLng(feature.geometry.coordinates.get(1),
+        feature.geometry.coordinates.get(0));
+    final Locale locale = Locale.US;
+    final CharSequence name = feature.properties.name;
+    final List<Integer> placeTypes = new ArrayList<>();
+    if (feature.properties.layer != null) {
+      placeTypes.addAll(layerConverter.getPlaceTypes(feature.properties.layer));
+    }
+    final LatLngBounds viewport = new LatLngBounds(latLng, latLng);
```
ecgreb: Currently the viewport returned is always a single point. Should we set some sensible default bounds that could work to display a venue on the map and maybe a few surrounding blocks?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceAutocompleteActivity.java
```diff
@@ -67,7 +68,9 @@
     }
 
     // TODO inject
-    PlaceDetailFetcher detailFetcher = new PeliasPlaceDetailFetcher();
+    Pelias pelias = new Pelias();
```
ecgreb: The `Pelias` instance passed into the `PeliasPlaceDetailFetcher` needs to be configured to use the Mapzen API key.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlacePickerActivity.java
```diff
@@ -46,7 +47,9 @@
     autocompleteView.setActivity(this);
 
     //TODO inject
-    presenter = new PlacePickerPresenterImpl(new PeliasPlaceDetailFetcher());
+    Pelias pelias = new Pelias();
```
ecgreb: The `Pelias` instance passed into the `PeliasPlaceDetailFetcher` needs to be configured to use the Mapzen API key.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
sarahlensing: Added info in javadocs and am throwing an exception now when those fields are attempted to be accessed. I'll also update the documentation PR.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureToPlaceConverter.java
sarahlensing: Good call. Looks like we had logic to do this in `PlaceAutocompletePresenter`...I've extracted this logic into a new class and am now using it in both places to generate a bounding box given a point.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceAutocompleteActivity.java
sarahlensing: Migrated to use the MapzenSearch obj.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlacePickerActivity.java
sarahlensing: Migrated to use a MapzenSearch obj

PR #288, opened 2017-01-31
Stylesheet updates and API key global vars

### LINKED ISSUE #114: Set tile API key in scene config and update stylesheets
It used to be that the only way to set a tile API key for tangram-es at runtime was to append it as a URL query to each tile request. This is no longer the case. In fact it is now preferred to set the API key for each data source as part of the scene configuration ([docs](https://mapzen.com/documentation/tangram/sources/#url_params)). As described in [this issue](https://github.com/tangrams/outdoor-style/issues/52#issuecomment-224087517), the append-directly-to-url method breaks down if, for example, vector tiles and terrain tiles require two different API keys. In our code, setting the key would look something like:

``` java
mapController.queueSceneUpdate("sources.osm.url_params", "{ api_key: ${MY_API_KEY} }");
mapController.applySceneUpdates();
```

ecgreb:
### Overview

Upgrades all bundled house stylesheets in the SDK to the latest release version. Upgrades Tangram to version 0.4.9-SNAPSHOT to support new scene update at map load time. Sets API key at map load time using stylesheet global variable syntax.

### Proposed Changes

* Upgrades stylesheet submodules to latest version using global vars and no default API key.
* Upgrades Tangram to version 0.4.9-SNAPSHOT to support new `SceneUpdate` syntax.
* Removes dynamic API key setting in `TileHttpHandler`.
* Sets Mapzen API key inside `getMapAsync(...)` using Tangram `SceneUpdate`.
* Sets `MapInitializer` dependencies using constructor injection rather than field injection.
* Updates calls to `mapController.queueSceneUpdate(...)` to use `SceneUpdate` param.
* Replaces OkHttp2 classes with OkHttp3 classes in `TileHttpHandler`.
* Adds SSL mocking utils for tests that depend on `HttpHandler` and `OkHttp3`.

Closes #114


ecgreb:
It turns out there is a much cleaner way to solve the testing issue. PowerMock is known to interact poorly with the `javax.net.ssl.*` package and so added an explicit ignore and removed the custom mocks even though they were so much fun to write 😜 

http://stackoverflow.com/questions/14654639/when-a-trustmanagerfactory-is-not-a-trustmanagerfactory-java

ecgreb:
@tallytalwar I purposely chose the one release before the latest for each stylesheet since this is the version with no API key in the scene file. The latest x.x.1 releases added back temporary API keys for 1 month (which we don't need on mobile).

PR #292, opened 2017-02-02
Refactor Gradle Properties

ecgreb:
### Overview

Cleans up gradle properties including version numbers and other fields used to generate the Maven POM. Synchronizes version number for all modules (core, sdk, places).

Updates API key gradle property for sample app and fixes a `NullPointerException` resulting from API key not getting properly set in places asmple app.

### Proposed Changes

* Extract common Gradle properties in root `gradle.properties` file including single version number for entire project.
* Update `build.gradle` files for sample applications to set `MAPZEN_API_KEY` value in `BuildConfig.java`.
* Update Circle CI sample app deploy scripts.
* Fixes NPE in PlaceDetailFetcher when no results are returned for reverse geocode. That happens when the API key is not properly set but can also theoretically occur for an area with low POI density.
* Changes Places API sample app name from "Mapzen Places API" to "Mapzen Places" to match SDK demo app which is titled "Mapzen SDK".
* Centralize API key management in new `MapzenManager` class.


mapzen-android-sdk/build.gradle
```diff
@@ -81,7 +81,7 @@ task verify(dependsOn: ['compileDebugSources',
                         'lint'])
 
 dependencies {
-  compile 'com.mapzen:mapzen-core:0.0.1-SNAPSHOT'
```
ecgreb: Up until now `core` and `places` were using version 0.0.1-SNAPSHOT and `sdk` was using version 1.3.0-SNAPSHOT (see core/gradle.properties). This pull request syncs all modules to use version 1.3.0-SNAPSHOT.

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasCallbackHandler.java
```diff
@@ -28,12 +26,17 @@
    * Creates a {@link Place} from a {@link Feature} with a name that matches the given title and
    * notifies the listener that a place has been successfully retrieved.
```
ecgreb: Good catch - updated

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasCallbackHandler.java
```diff
@@ -28,12 +26,17 @@
...
    * @param title
-   * @param features
+   * @param response
    * @param listener
    */
-  public void handleSuccess(String title, List<Feature> features,
+  void handleSuccess(String title, Response<Result> response,
```
ecgreb: It should be fine. This is an internal callback only. We try to make everything in the `com.mapzen.places.api.internal` package protected so it will not show up in the public API.

Issue #279, opened 2017-01-26
Deploy Places 1.3.0-rc1 Artifact


sarahlensing:
Deploy to Maven Central

ecgreb:
http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.mapzen%22

PR #298, opened 2017-02-08
Improve build performance

### LINKED ISSUE #297: Circle CI out of memory error

ecgreb:
### Overview

Improves build performance on Circle CI. Reduces build time by 1-2 min for dev builds and 4-5 min for full master builds (includes sample apps). Prevents OOM errors on Circle CI container.

### Proposed Changes

* Updates Gradle options to limit memory use and dump heap on OOM error.
* Runs unit tests against debug build only and reduces console output.
* Upgrades Gradle to version 3.3 and gradle-android-maven-plugin to version 1.5.
* Adds `--configure-on-demand` and `-PdisablePreDex` options to circle.yml.
* Runs tests for each module in parallel using 3 Circle CI containers instead of 1.
* Archives sample app APKs on Circle CI and deploys to AWS on release build only.
* Publishes documentation to Mapzen website on release build only.
* Installs `core` module to local .m2 repo before resolving dependencies.
* Disables Javadoc tasks for all modules. May need to be re-enabled for production AARs.

Fixes #297

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,8 +16,6 @@ dependencies:
     - echo y | android update sdk --all --no-ui --force --filter "build-tools-25.0.2"
     - echo y | android update sdk --all --no-ui --force --filter "android-25"
     - echo y | android update sdk --all --no-ui --force --filter "extra-android-m2repository"
-    - sudo apt-get update; sudo apt-get install s3cmd
-    - printf "[default]\naccess_key = $S3_ACCESS_KEY\n secret_key = $S3_SECRET_KEY" > ~/.s3cfg
 
 test:
   override:
@@ -29,8 +27,7 @@ deployment:
     branch: master
     commands:
       - ./gradlew uploadArchives -PsonatypeUsername=$SONATYPE_NEXUS_SNAPSHOTS_USERNAME -PsonatypePassword=$SONATYPE_NEXUS_SNAPSHOTS_PASSWORD
-      - scripts/deploy-android-sdk-sample-app.sh
-      - scripts/deploy-places-api-sample-app.sh
       - ./gradlew aarSize countReleaseDexMethods permissions mapzen-android-sdk:dependencies --configuration compile
+      - scripts/deploy-samples.sh
       - pip install 'Circle-Tickler == 1.0.1'
       - tickle-circle mapzen documentation master $CIRCLE_TOKEN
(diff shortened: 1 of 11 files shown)


tallytalwar:
Starting reviewing this now.

circle.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
 
 test:
   override:
-    - ./gradlew clean :core:install :core:verify :mapzen-android-sdk:verify :mapzen-places-api:verify
+    - ./gradlew clean :core:install :core:verify :mapzen-android-sdk:verify :mapzen-places-api:verify --configure-on-demand
```
tallytalwar: 👍 Nicely documented in gradle docs (which sometimes are very convoluted, at least for me): https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:configuration_on_demand

circle.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
 
 test:
   override:
-    - ./gradlew clean :core:install :core:verify :mapzen-android-sdk:verify :mapzen-places-api:verify --configure-on-demand
+    - ./gradlew clean :core:install :core:verify :mapzen-android-sdk:verify :mapzen-places-api:verify --configure-on-demand -PdisablePreDex
```
tallytalwar: The only information I found for this is its recommended to use `-PdisablePreDex` when doing a one time build (typical with CI env.). But no finer details behind the reasoning. Can you point me to a reference/explain? Thanks

scripts/publish-docs.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+#
+# Triggers mapzen docs build to publish to https://mapzen.com/documentation/.
+
+if [[ ${PERFORM_RELEASE} ]]
```
tallytalwar: naive question ... when does `PERFORM_RELEASE` get set?

circle.yml
ecgreb: Using `-PdisablePreDex` was recommended in [this Circle CI issue](https://discuss.circleci.com/t/java-lang-outofmemoryerror-permgen-space-in-android-gradle-build/3652).

scripts/publish-docs.sh
ecgreb: Haha it does not yet! :) I will be adding a script to perform the release using Circle's API similar to [this one we use for Eraser Map](https://github.com/mapzen/eraser-map/blob/master/scripts/perform-release.sh) as part of https://github.com/mapzen/android/issues/279.

PR #299, opened 2017-02-08
Use MapzenManager only to set API keys in code

### LINKED ISSUE #293: Components should fail fast if not API key has been set
### Description

Any component that attempts to make a request against a Mapzen service should throw an `IllegalStateExceptoin` if a Mapzen API has not been declared as a string resource or set in code.

The exception thrown should include an error message that directs the developer to documentation on how to register for a Mapzen API key and inject it into the application.

### Steps to Reproduce

Run the SDK or Places demo app. Do not set an API as a gradle property or in `mapzen.xml`. Search and routing features will not work properly but not explicit exception is thrown.

### Mapzen SDK & Android Version

Mapzen 1.3.0-SNAPSHOT
Android 7.1.1

ecgreb:
### Overview

Removes ability to initialize services with a component specific API key. All API keys should be provided via `MapzenManager#setApiKey(String)` or a string resource in `mapzen.xml`.

Throws an exception before any requests are made to Mapzen services if an API key has not been provided.

### Proposed Changes

* Throws exception if MapzenManager has not been initialized with a valid API key.
* Prevents setting vector tiles API key directly on `MapView` via `getMapAsync(...)`.
* Removes MapzenSearch constructor with API key.
* Removes MapzenRouter constructor with API key.
* Provides Mapzen API key via MapzenManager to all SDK samples activities.

Fixes #293


core/src/main/java/com/mapzen/android/core/MapzenManager.java
```diff
@@ -48,16 +49,24 @@ public static MapzenManager instance(Context context) {
...
+      int id = resources.getIdentifier(API_KEY_RES_NAME, API_KEY_RES_TYPE,
+          context.getPackageName());
+      if (id > 0) {
+        apiKey = resources.getString(id);
+      }
     }
   }
 
   /**
    * Returns the currently active API key stored by this class.
```
tallytalwar: Can you update the comments here to this throws an exception if a valid key is not already set.

core/src/test/java/com/mapzen/android/search/MapzenSearchTest.java
```diff
@@ -18,16 +21,28 @@
 import retrofit2.Response;
 
 public class MapzenSearchTest {
+  private Context context;
+  private Pelias pelias;
+  private MapzenSearch search;
+
+  @Before public void setUp() throws Exception {
```
tallytalwar: nitpick, probably `setup` instead of camel-case `setUp`, unless there is some convention here.

tallytalwar: totally fine either way :D.

ecgreb: This naming convention is actually a legacy of JUnit 3 when the `@Before` annotation did not exist and the pre-test method had to be named `setUp()` with camel case in order to be picked up by the test runner.

Fun fact is before JUnit annotations existed all test methods had to start with the word `test` too. For example `testNotNull()`.

While starting each test method name with the word `test` has been largely abandoned in JUnit 4 now that we have the `@Test` annotation most developers still adhere to the `@Before setUp()` and `@After tearDown()` naming conventions.

core/src/main/java/com/mapzen/android/core/MapzenManager.java
ecgreb: Good catch! Added in https://github.com/mapzen/android/pull/299/commits/2675af43ae1a6f99cdd20f1dc1dbe9185ae20f76.

core/src/test/java/com/mapzen/android/search/MapzenSearchTest.java
tallytalwar: Ohh.. good to know. The things I learn :D.

PR #302, opened 2017-02-09
Updates place picker fetch details logic

### LINKED ISSUE #295: Improve place detail fetch algorithm
Should not be string matching title

ecgreb:
### Overview

Replaces reverse geocode and string match on title with a constructed Pelias gid and /place endpoint when a map POI is selected in the place picker widget.

### Proposed Changes

If tile properties contains an `area` for the selected POI then the Pelias gid is constructed using "openstreetmap:venue:**way**:[ID]" otherwise uses "openstreetmap:venue:**node**:[ID]" to fetch details from the /place endpoint.

Fixes #295

### DIFF
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceDetailFetcher.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceDetailFetcher.java
@@ -1,7 +1,5 @@
 package com.mapzen.places.api.internal;
 
-import com.mapzen.tangram.LngLat;
-
 import java.util.Map;
 
 /**
@@ -9,21 +7,26 @@ import java.util.Map;
  */
 interface PlaceDetailFetcher {
 
-  /**
-   * Called when details for a place selected from the map should be retrieved. Currently, Pelias is
-   * used as the underlying data source but this will soon be migrated to WOF.
-   * @param coordinates
-   * @param properties
-   * @param listener
-   */
-  void fetchDetails(LngLat coordinates, Map<String, String> properties,
-      OnPlaceDetailsFetchedListener listener);
+  String PROP_ID = "id";
+  String PROP_AREA = "area";
+
+  String PELIAS_GID_BASE = "openstreetmap:venue:";
+  String PELIAS_GID_NODE = "node:";
+  String PELIAS_GID_WAY = "way:";
 
   /**
-   * Called when details for a place selected from autocomplete should be retrieved. Currently,
-   * Pelias is used as the underlying data source but this will soon be migrated to WOF.
-   * @param gid
-   * @param listener
+   * Called when details for a place selected from the map should be retrieved.
+   *
+   * @param properties map properties for the given feature.
+   * @param listener callback to be notified of success or failure.
+   */
+  void fetchDetails(Map<String, String> properties, OnPlaceDetailsFetchedListener listener);
+
+  /**
+   * Called when details for a place selected from autocomplete should be retrieved.
+   *
+   * @param gid global ID for the given feature
+   * @param listener callback to be notified of success or failure.
    */
   void fetchDetails(String gid, OnPlaceDetailsFetchedListener listener);
 }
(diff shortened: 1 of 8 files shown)


mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceDetailFetcher.java
```diff
@@ -1,29 +1,32 @@
...
 
 /**
  * Interface for fetching details about a place.
  */
 interface PlaceDetailFetcher {
 
+  String PROP_ID = "id";
+  String PROP_AREA = "area";
+
+  String PELIAS_GID_BASE = "openstreetmap:venue:";
```
tallytalwar: just confirming if we know that the pelias gid for the place end point always is a `venue`?

mapzen-places-api/src/test/java/com/mapzen/places/api/internal/PeliasPlaceDetailFetcherTest.java
```diff
@@ -1,44 +1,60 @@
...
+  @Test public void fetchDetails_gid_shouldCallPelias() throws Exception {
     String gid = "123";
     fetcher.fetchDetails(gid, null);
     verify(pelias).place(eq(gid), any(Callback.class));
   }
+
+  @Test public void fetchDetails_props_shouldConstructGidWithWayIfHasArea() throws Exception {
+    Map props = new HashMap();
+    props.put(PROP_ID, "123.000");
+    props.put(PROP_AREA, "456.000");
```
tallytalwar: Based on http://wiki.openstreetmap.org/wiki/Area, area property could also be a string like "yes". Maybe we add another test for `yes` area value. Though in the main code its correctly checking for the existence of the property, so should be all good.

tallytalwar: Also from the above link only, it seems `highway=footway` `closed-ways` are not assumed to have the area property, how do these get distinguished between a node or a way?

mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PlaceDetailFetcher.java
ecgreb: In all the cases I tested for interactive icons that show on the map the answer was yes.

mapzen-places-api/src/test/java/com/mapzen/places/api/internal/PeliasPlaceDetailFetcherTest.java
ecgreb: Good catch I added a test case for `area=yes` to be sure the production code does not choke on it.

https://github.com/mapzen/android/pull/302/commits/f89d94e5ca8f690680201bd1457f203347d9ec4d

In the mobile version of the map roads should not be interactive so I think we are safe in the case of something like `highway=footway`?

tallytalwar: 👍

ecgreb:
OSM ID inconsistencies are a real concern if a bit of an edge case.

At worst they are a temporary problem until the search index is rebuilt and picks up the new ID (usually once a week).

Regarding WOF we are still a ways off from being able to integrate WOF IDs since they are not currently contained in the tile metadata.

We could fallback to the old method (reverse geo + title match) when the new method returns no results but my hit would be let's see how often it actually occurs before we over engineer a solution for something that is likely a temporary edge case.

tallytalwar:
Sounds good. All bases covered :D.
Hitting the green button.

bcamper:
@ecgreb did you review this logic with @nvkelso? I have some concern that expecting properties like `area` to imply way vs. node could be brittle, especially as the tile schema evolves.

ecgreb:
@bcamper I did go over the logic with @nvkelso. Actually I believe he used this same logic to fix a bug with the OSM edit links generated on http://tangrams.github.io/bubble-wrap/.

You are correct with this approach we would need to coordinate any breaking changes in the tile schema with an SDK update. But we would already need to do that since the SDK also depends on other tile properties besides `area`.

PR #307, opened 2017-02-17
Map position should be set with the position of the searched feature

tallytalwar:


ecgreb:
Thanks @tallytalwar !

PR #309, opened 2017-02-21
One more try using project dependency for core module

ecgreb:
### Overview

Use a project dependency for `core` in the sdk and places modules.

### Proposed Changes

Replace core module remote dependency with project dependency in sdk and places modules. This will not replace needing to deploy the `mapzen-core` module to maven central but rather should make local development workflow simpler and resolve build issues on Circle CI when version number is updated.

Also transitive dependencies seem to get resolved correctly for the core module when a snapshot is deployed locally now that the `gradle-release` plugin config has been updated.

### DIFF
--- a/mapzen-android-sdk/build.gradle
+++ b/mapzen-android-sdk/build.gradle
@@ -57,7 +57,7 @@ task checkstyle(type: Checkstyle) {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


ecgreb:
Merging to test snapshot deployment from master.

PR #317, opened 2017-03-01
reorg docs & add places

kkowalsky:
### Overview

Changing some of the docs around to be more contained and allow for the Places API. 

### Proposed Changes
- merges polygons, polylines, markers into add-features.md
- merges current-location and location-services into location-services.md
- Created places.md for Places API


docs/installation-places.md
```diff
@@ -1,35 +0,0 @@
-# Installation
```
ecgreb: Where did the installation docs for places get moved to or did they maybe get lost in the shuffle?

These should still be included since places is distributed as its own library (AAR) but perhaps we could combine these with the SDK install docs into a single `installation.md`?

kkowalsky: I deleted the places installation because it seemed identical? I can add a separate section to installation about installing places as well.

ecgreb: True there is a lot that is the same the main difference is the artifact name `mapzen-android-sdk` vs. `mapzen-places-api` but maybe some of the text only needs to be in there once.

kkowalsky: ahhh, i'll add a section in places.md and installation.md about installing Places

ecgreb:
👍

PR #337, opened 2017-03-18
Fixed crash when mapzenMap field was not set in MapView instance

rwrx:
### Overview
Fixed crash when mapzenMap field was not set in MapView instance. This caused crash in my app when `MapzenMap` wasn't initialized because my app uses Google Maps API and Mapzen Android SDK and only one of them can be active at a time causing that one of them won't be initialized at app startup.

### Proposed Changes
Add null pointer check to `onDestroy()` method in `MapView` class.


ecgreb:
Thanks @rwrx !

PR #347, opened 2017-03-31
Fixed some Lost memory leaks

rwrx:
### Overview
With help of Leak Canary https://github.com/square/leakcanary I have found some memory leaks inside Mapzen Android SDK in OverlayManager.

### Proposed Changes
I am not sure whether creating new LostApiClient inside OverlayManager constructor instead of using sharedClient is a correct solution, but it is very dangerous to set Context into static field variable as it will always leak when not handled properly.

### DIFF
--- a/core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
+++ b/core/src/main/java/com/mapzen/android/graphics/OverlayManager.java
@@ -152,7 +152,7 @@ public class OverlayManager implements TouchInput.PanResponder, TouchInput.Rotat
   OverlayManager(MapView mapView, MapController mapController, MapDataManager mapDataManager,
       MapStateManager mapStateManager, LostApiClient lostApiClient) {
     if (lostApiClient == null) {
-      lostApiClient = LocationFactory.sharedClient(mapView.getContext(), connectionCallbacks);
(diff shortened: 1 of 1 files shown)


sarahlensing:
We could have also used the application context for this shared client but I think its cleaner to just create a new client as you have done. Besides, it looks like if different connection callbacks were passed in, they weren't added to the shared client.

Thanks for your PR @rwrx

ecgreb:
I'm concerned that creating a new Lost client in the `OverlayManager` constructor each time will lead to thrashing of the location service if it is disconnected and reconnected each time the device is rotated. More testing is needed to confirm if that is actually the case.

I like the idea of using the application context since that is scoped to the lifecycle of the app and not the activity. This is also how the Lost client is managed in the Mapbox SDK.

https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java#L39-L45

As an additional safeguard, the Mapbox `LostLocationEngine` only holds a reference to the `Context` using a `WeakReference`. This way if the `LostLocationEngine` has the only reference to the `Context` then it will be released and garbage collected.

rwrx:
As I wrote in Lost project, yes, think that, WeakReference can help with leaking Context. Is there some way, I can help with resolving this?

PR #351, opened 2017-04-18
Use https instead of ssh for tron submodule

tallytalwar:
Small fixup


ecgreb:
thanks @tallytalwar !

PR #361, opened 2017-05-05
Update tangram to use 0.5.2

tallytalwar:
- has a bunch of bug fixes for older android devices

Not updated to 0.6.2, has thats a major upgrade and will require more testing and probably more changes. Will create an issue for that though.


msmollin:
Per resulting discussion I think we're gonna hold off doing this upgrade and instead target 0.6.2 / 0.7 for inclusion.

PR #371, opened 2017-05-15
Improve Style Switching

### LINKED ISSUE #363: Walkabout style missing api key
### Description
The api key is not properly appended to tile requests and we reach `MapController`'s L1122 IOException.

The sdk is on an old version of Walkabout (3.0.0 from Jan 19) and should probably be updated.

sarahlensing:
### Overview
This PR updates `MapzenMap#setStyle` so that it reconfigures global scene updates (such as api key and locale) when a new style is loaded

Closes #363


sarahlensing:
@tallytalwar yes it does

PR #389, opened 2017-05-30
Return Place from Feature if no gid present

### LINKED ISSUE #388: Update places api for optional gid
If there is no `gid`, instead of returning no `Place`, we should return a `Place` constructed from the info returned by the autocomplete endpoint

sarahlensing:
### Overview
To handle optional `gid` values, this code updates the `Places` API to simply return a `Place` object from the current `Feature` instead of trying to fetch details given an empty `gid`

### Proposed Changes
- Refactors code to create a `Place` and details string into the same `PeliasFeatureConverter` object to centralize logic
- Updates documentation
- Adds more test coverage for optional `gid`

Closes #388

### DIFF
--- a/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureConverter.java
+++ b/mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureConverter.java
@@ -10,10 +10,10 @@ import java.util.List;
 import java.util.Locale;
 
 /**
- * Handles converting Pelias {@link Feature} objects into {@link Place} objects for
- * {@link PeliasCallbackHandler}.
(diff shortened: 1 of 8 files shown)


mapzen-places-api/src/main/java/com/mapzen/places/api/internal/PeliasFeatureConverter.java
```diff
@@ -59,4 +59,26 @@ Place getFetchedPlace(Feature feature) {
...
+   */
+  String getDetails(Feature feature) {
+    String title = feature.properties.name;
+    return getDetails(feature, title);
+  }
+
+  /**
+   * Constructs a detail string from a {@link Feature} and title.
+   * @param feature
+   * @param title
```
tallytalwar: can we elaborate on `title` here?

Issue #426, opened 2017-08-09
Support Tangram Marker polygon and polyline styles


sarahlensing:
Currently only the "points" style is officially supported in the SDK. We should add support for the other two styles Tangram supports. We should also add support for some "system marker" types for common stylesheet elements like the sdk polygon and polyline styles (like we do in the ios sdk).

Issue #343, opened 2017-03-24
Marker size should be configurable without setting entire style string


ecgreb:
Developers should be able to icon size using `MarkerOptions#size(int width, int height)` to set width and height pixel values contained in the style string.

sarahlensing:
Closed via https://github.com/mapzen/android/pull/427

Issue #342, opened 2017-03-24
Create custom bitmap marker using a Drawable object


ecgreb:
Currently only resource ID is supported for custom marker icons. Add support for any `Drawable` object (or bitmap).

sarahlensing:
Closed via https://github.com/mapzen/android/pull/427

Issue #381, opened 2017-05-22
Convert current location gem and search pin annotations to tangram marker interface


msmollin:
After discussion with @tallytalwar we should convert the current location gem, the driving icon, and the search pin annotations to use the MapData objects.

tallytalwar:
Search pins can use Marker interface, maybe any polygon or polyline example could use MapData (clientdatasource) interface.

msmollin:
Bah yeah I got it backwards in the ticket description. Edited and fixed.

PR #427, opened 2017-08-09
Expand BitmapMarker API

### LINKED ISSUE #344: Modify marker position, icon, and size after creation
After a maker is created using `MapzenMap#addMarker(MarkerOptions)` a developer should be able to modify the position, icon, size, and visibility of the `BitmapMarker` created.

sarahlensing:
### Overview
This PR adds support for several `BitmapMarker` APIs covering most of the Tangram `Marker` API. It also adds support for additional properties by internally converting some properties (size, interactivity) into a style string. 

### Proposed Changes
Adds a new class, `StyleStringConverter` to handle generating the correct style string for `BitmapMarker`s. Updates and adds test coverage for new functionality. Removes the raw style string property from `MarkerOptions`.

[Future work](https://github.com/mapzen/android/issues/426) will add remaining Tangram `Marker` API support (polygons and polylines). 

Closes #344, #343, #342

### DIFF
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -1,7 +1,5 @@
 package com.mapzen.android.graphics;
 
-import com.mapzen.android.graphics.internal.StyleStringGenerator;
-
 import javax.inject.Singleton;
 
 import dagger.Module;
@@ -37,11 +35,4 @@ import dagger.Provides;
     return new SceneUpdateManager();
   }
 
-  /**
-   * Returns the object used to generate the underlying style string.
-   * @return
-   */
-  @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
-    return new StyleStringGenerator();
-  }
 }
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -1,5 +1,7 @@
 package com.mapzen.android.graphics;
 
+import com.mapzen.android.graphics.internal.StyleStringGenerator;
+
 import javax.inject.Singleton;
 
 import dagger.Module;
@@ -33,5 +35,13 @@ import dagger.Provides;
    */
   @Provides @Singleton public SceneUpdateManager providesSceneUpdateManager() {
     return new SceneUpdateManager();
+  }
+
+  /**
+   * Returns the object used to generate the underlying style string.
+   * @return
+   */
+  @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
+    return new StyleStringGenerator();
   }
 }
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -34,5 +34,4 @@ import dagger.Provides;
   @Provides @Singleton public SceneUpdateManager providesSceneUpdateManager() {
     return new SceneUpdateManager();
   }
-
 }
--- a/core/src/main/java/com/mapzen/android/graphics/MapzenMap.java
+++ b/core/src/main/java/com/mapzen/android/graphics/MapzenMap.java
@@ -50,7 +50,6 @@ public class MapzenMap {
   private final MarkerManager markerManager;
   private final SceneUpdateManager sceneUpdateManager;
   private final MapzenManager mapzenManager;
-  private final StyleStringGenerator styleStringGenerator;
   private Locale locale;
 
   boolean pickFeatureOnSingleTapConfirmed = false;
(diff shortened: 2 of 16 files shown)


tallytalwar:
Looking at this now.

core/src/main/java/com/mapzen/android/graphics/internal/StyleStringGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  public void setInteractive(boolean interactive) {
+    this.interactive = interactive;
+  }
+
+  /**
+   * Sets the hex value for background color to be used.
+   * @param hex
+   */
+  public void setBackgroundColor(String hex) {
```
tallytalwar: Is this marker color? Maybe `BackgroundColor` is not an apt name here.

core/src/main/java/com/mapzen/android/graphics/model/BitmapMarker.java
```diff
@@ -26,4 +35,114 @@ public BitmapMarker(MarkerManager markerManager, Marker tangramMarker) {
...
+   * @param duration
+   * @param easeType
+   */
+  public void setPosition(LngLat position, int duration, EaseType easeType) {
+    this.tangramMarker.setPointEased(position, duration,
+        EaseTypeConverter.EASE_TYPE_TO_MAP_CONTROLLER_EASE_TYPE.get(easeType));
+  }
+
+  /**
+   * Sets the resource id displayed as the marker's icon.
```
tallytalwar: `drawable` resource id.

core/src/main/java/com/mapzen/android/graphics/model/Marker.java
```diff
@@ -3,7 +3,7 @@
 import com.mapzen.tangram.LngLat;
 
 /**
- * Represents a pin marker on a map.
+ * Represents a pin marker on a map backed by {@link MapData} objects.
```
tallytalwar: Broken link here! I think might have to import `com.mapzen.tangram.MapData`.

core/src/main/java/com/mapzen/android/graphics/MapzenMap.java
```diff
@@ -649,7 +643,8 @@ public void onMarkerPick(final MarkerPickResult markerPickResult, final float po
         mapView.post(new Runnable() {
           @Override public void run() {
             if (markerPickResult != null) {
-              listener.onMarkerPick(new BitmapMarker(markerManager, markerPickResult.getMarker()));
+              listener.onMarkerPick(new BitmapMarker(markerManager, markerPickResult.getMarker(),
```
tallytalwar: Comment on the documentation above for `setMarkerPickListener`.

Because of the distinction between a `Marker` (backed by `Tangram.MapData`) and a `BitmapMarker` (backed by `Tangram.Marker`), I think the documentation should clarify this to be a `BitmapMarker` and not a `Marker`.

tallytalwar: Also the `FeaturePickListener` is the one which handles picking of `Markers` (the one backed by `Tangram.MapData`)

core/src/main/java/com/mapzen/android/graphics/internal/StyleStringGenerator.java
sarahlensing: Renamed to `setColor`

core/src/main/java/com/mapzen/android/graphics/model/BitmapMarker.java
sarahlensing: Updated

core/src/main/java/com/mapzen/android/graphics/model/Marker.java
sarahlensing: Updated

core/src/main/java/com/mapzen/android/graphics/MapzenMap.java
sarahlensing: Updated. Thanks for closely looking over the javadocs! 😄

tallytalwar: Ok this is a bit confusing to me, why create a new `BitmapMarker` in the listener interface here? I might not be fully understanding this here.

Looks like this was being done originally here: https://github.com/mapzen/android/pull/340/files#diff-acea21ee9a7821af52798402334ac2e1R638

sarahlensing: This should be updated to use code to map the Tangram `Marker` to the existing `BitmapMarker`. Created a ticket to do this in future work: https://github.com/mapzen/android/issues/429

core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
```diff
@@ -37,11 +35,4 @@
     return new SceneUpdateManager();
   }
 
-  /**
-   * Returns the object used to generate the underlying style string.
-   * @return
-   */
-  @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
```
sarahlensing: It can't be a singleton because it keeps state related to an individual marker. I could update it to not save state and have it be a singleton again but I think I prefer to leave it as is because I imagine using it to restore markers when a new style is loaded https://github.com/mapzen/android/issues/348

sarahlensing:
Good catch, we were planning to make those updates as part of this ticket: https://github.com/mapzen/android/issues/381

Issue #348, opened 2017-03-31
Restore custom markers on rotation


sarahlensing:
If the map is set to persist data, it should restore custom bitmap markers added to the map

PR #428, opened 2017-08-10
Tangram 0.8.0-RC2

### LINKED ISSUE #393: Upgrade to Tangram 0.8.0

sarahlensing:
### Overview
Upgrades SDK to use the latest Tangram release candidate. ~This has breaking API changes to `MapzenMap` as `queueSceneUpdate ` and `applySceneUpdates` have been removed.~ This deprecates  `queueSceneUpdate ` and `applySceneUpdates`. The new `updateSceneAsync ` method has been left out of the public interface for now because the primary use cases for the Android SDK should ensure that users don't need to access this method directly. If they want to use it, they still can via `MapzenMap#getMapController#updateSceneAsync`.

### Proposed Changes
Uses new `SceneLoadListener` callback to create and return a new `MapzenMap` via the SDK `onMapReady` callback. Updates the code to ensure that this method is only invoked once. These changes also improve test coverage.

[Future work](https://github.com/mapzen/android/issues/348) will add marker restoration support

Closes #393


core/src/main/java/com/mapzen/android/graphics/MapInitializer.java
```diff
@@ -90,15 +91,19 @@ private void loadMap(final MapView mapView, String sceneFile, final OnMapReadyCa
     final List<SceneUpdate> sceneUpdates = sceneUpdateManager.getUpdatesFor(apiKey, locale,
         mapStateManager.isTransitOverlayEnabled(), mapStateManager.isBikeOverlayEnabled(),
         mapStateManager.isPathOverlayEnabled());
```
sarahlensing: I'm glad you brought this up and think it would be nice to provide more global options like this. Created tickets in the iOS and Android SDKs for future work to complete this
https://github.com/mapzen/android/issues/431
https://github.com/mapzen/ios/issues/355

core/src/main/java/com/mapzen/android/graphics/MapInitializer.java
```diff
@@ -90,15 +89,13 @@ private void loadMap(final MapView mapView, String sceneFile, final OnMapReadyCa
...
-    getTangramView(mapView).getMapAsync(new com.mapzen.tangram.MapView.OnMapReadyCallback() {
-      @Override public void onMapReady(MapController mapController) {
-        mapController.setHttpHandler(mapzenMapHttpHandler.httpHandler());
-        MapzenManager mapzenManager = MapzenManager.instance(mapView.getContext());
-        callback.onMapReady(
-            new MapzenMap(mapView, mapController, new OverlayManager(mapView, mapController,
-                mapDataManager, mapStateManager), mapStateManager, new LabelPickHandler(mapView),
-                new MarkerManager(mapController), sceneUpdateManager, locale, mapzenManager));
+    MapController controller = mapView.getTangramMapView().getMap(
```
sarahlensing: The reason I set the `mapController`'s `SceneLoadListener` to `null` is because: 
1. After the initial scene loads, the SDK considers the map "ready" and doesn't want to invoke the `OnMapReadyCallback` for other calls to change the scene etc (looking at your comment below, I could also update this to check the sceneId)
2. The SDK provides type safety around scene updates (https://github.com/mapzen/android/blob/master/core/src/main/java/com/mapzen/android/graphics/SceneUpdateManager.java) and is tested to ensure the calls to sceneUpdateAsync don't produce errors. I know there has been talk about supporting external scenes but at the moment, the SDK doesn't have first class support for this case. If developers care about errors we expect them to access the `MapController` directly and set a `SceneLoadListener` on this object. There will be future discussion around this expansion of the SDK though.

Does this reasoning make sense, are there things I'm not considering properly?

On a related note, I was planning on setting a `SceneLoadListener` as part of the marker restoration work here though (https://github.com/mapzen/android/issues/348).

core/src/main/java/com/mapzen/android/graphics/MapInitializer.java
```diff
@@ -90,15 +89,13 @@ private void loadMap(final MapView mapView, String sceneFile, final OnMapReadyCa
...
+        new MapController.SceneLoadListener() {
+      @Override public void onSceneReady(int sceneId, SceneError sceneError) {
+        mapReadyInitializer.onMapReady(mapView, mapzenMapHttpHandler, callback, mapDataManager,
```
sarahlensing: Because of the way our APIs are setup I don't think this is needed.

The flow the user goes through is either:
1. `MapView#getMapAsync(OnMapReadyCallback)` or `MapFragment#getMapAsync(OnMapReadyCallback)`
2. `other calls to the map after receiving it in OnMapReadyCallback`

We wrap all calls to `MapController` so it is only after receiving this callback that the user has a `MapzenMap` and can make additional calls to `MapController`. This means that the first call we get to `onSceneReady` is guaranteed to be the one we care about because it is in response to the only call we have made to `MapController` (unless the user has accessed `MapController` directly and made calls to it. But this case can lead to breaking other parts of our API that we don't support).

PR #433, opened 2017-08-22
Restore Markers

### LINKED ISSUE #348: Restore custom markers on rotation

sarahlensing:
### Overview
This PR uses `SceneLoadListener` to restore markers on scene updates, including scene updates which occur because of orientation changes. This code doesn't check `sceneId` in the listener callback because we always want to restore markers when the scene is ready after any update. It also doesn't check `sceneError` because the sdk doesn't have first class support for loading custom scenes and ensures that all house styles work with the apis we provide.

### Proposed Changes
- Adds `SceneLoadListener` to `MapzenMap`
- Adds restoration handling to `MarkerManager`
- Updates `MarkerOptions` to include all `BitmapMarker` properties
- Updates `BitmapMarker` to provide property getters
- Migrates `StyleStringGenerator` and `MarkerManager` to be singletons
- Updates and adds test coverage
- Updates custom marker example in sample app to show marker restoration on scene load changes

Closes #348

### DIFF
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -2,6 +2,7 @@ package com.mapzen.android.graphics;
 
 import com.mapzen.android.graphics.internal.StyleStringGenerator;
 import com.mapzen.android.graphics.model.BitmapMarkerFactory;
+import com.mapzen.android.graphics.model.MarkerManager;
 
 import javax.inject.Singleton;
 
@@ -52,5 +53,14 @@ import dagger.Provides;
    */
   @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
     return new StyleStringGenerator();
+  }
+
+  /**
+   * Returns the object used to manager markers.
+   * {@link com.mapzen.tangram.Marker}.
+   */
+  @Provides @Singleton public MarkerManager providesMarkerManager(
+      BitmapMarkerFactory bitmapMarkerFactory, StyleStringGenerator styleStringGenerator) {
+    return new MarkerManager(bitmapMarkerFactory, styleStringGenerator);
   }
 }
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -1,5 +1,8 @@
 package com.mapzen.android.graphics;
 
+import com.mapzen.android.graphics.internal.StyleStringGenerator;
+import com.mapzen.android.graphics.model.BitmapMarkerFactory;
+
 import javax.inject.Singleton;
 
 import dagger.Module;
@@ -33,5 +36,21 @@ import dagger.Provides;
    */
   @Provides @Singleton public SceneUpdateManager providesSceneUpdateManager() {
     return new SceneUpdateManager();
+  }
+
+  /**
+   * Returns the object used create {@link com.mapzen.android.graphics.model.BitmapMarker}s.
+   * @return
+   */
+  @Provides @Singleton public BitmapMarkerFactory providesBitmapMarkerFactory() {
+    return new BitmapMarkerFactory();
+  }
+
+  /**
+   * Returns the object used to generate a style string for a Tangram
+   * {@link com.mapzen.tangram.Marker}.
+   */
+  @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
+    return new StyleStringGenerator();
   }
 }
--- a/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
+++ b/core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
@@ -2,7 +2,7 @@ package com.mapzen.android.graphics;
 
 import com.mapzen.android.graphics.internal.StyleStringGenerator;
 import com.mapzen.android.graphics.model.BitmapMarkerFactory;
-import com.mapzen.android.graphics.model.MarkerManager;
+import com.mapzen.android.graphics.model.BitmapMarkerManager;
 
 import javax.inject.Singleton;
 
@@ -59,8 +59,8 @@ import dagger.Provides;
    * Returns the object used to manager markers.
    * {@link com.mapzen.tangram.Marker}.
    */
-  @Provides @Singleton public MarkerManager providesMarkerManager(
+  @Provides @Singleton public BitmapMarkerManager providesBitmapMarkerManager(
       BitmapMarkerFactory bitmapMarkerFactory, StyleStringGenerator styleStringGenerator) {
-    return new MarkerManager(bitmapMarkerFactory, styleStringGenerator);
+    return new BitmapMarkerManager(bitmapMarkerFactory, styleStringGenerator);
   }
 }
--- a/core/src/main/java/com/mapzen/android/graphics/model/BitmapMarkerFactory.java
+++ b/core/src/main/java/com/mapzen/android/graphics/model/BitmapMarkerFactory.java
@@ -1,0 +1,21 @@
+package com.mapzen.android.graphics.model;
+
+import com.mapzen.android.graphics.internal.StyleStringGenerator;
+
+/**
+ * Creates {@link BitmapMarker} objects.
+ */
+public class BitmapMarkerFactory {
+
+  /**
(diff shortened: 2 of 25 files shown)


core/src/main/java/com/mapzen/android/graphics/model/MarkerOptions.java
```diff
@@ -5,6 +5,7 @@
 
 import android.graphics.drawable.Drawable;
 
+
```
tallytalwar: Is this class only used for `BitmapMarkers` or also for `Markers` (backed by `MapData`)? If so, we should clarify the naming.

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
```diff
@@ -27,24 +45,61 @@ public MarkerManager(MapController mapController) {
...
-    } else {
-      marker.setDrawable(markerOptions.getIcon());
-    }
-    StyleStringGenerator styleStringGenerator = new StyleStringGenerator();
-    styleStringGenerator.setSize(markerOptions.getWidth(), markerOptions.getHeight());
-    marker.setStylingFromString(styleStringGenerator.getStyleString());
-    return new BitmapMarker(this, marker, styleStringGenerator);
+    BitmapMarker bitmapMarker = bitmapMarkerFactory.createMarker(this, marker,
+        styleStringGenerator);
+    configureMarker(bitmapMarker, markerOptions.getPosition(), markerOptions.getIconDrawable(),
```
tallytalwar: Instead of passing every property, can we not pass the markerOptions explicitly?

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
```diff
@@ -27,24 +45,61 @@ public MarkerManager(MapController mapController) {
...
+        markerOptions.getIcon(), markerOptions.getWidth(),
+        markerOptions.getHeight(), markerOptions.isInteractive(), markerOptions.getColorHex(),
+        markerOptions.getColorInt(), markerOptions.isVisible(), markerOptions.getDrawOrder(),
+        markerOptions.getUserData());
+    restorableMarkers.add(bitmapMarker);
```
tallytalwar: I could be missing something here, but this should be synchronized right? As a when a new scene is loaded its listener should have have a synchronized set of markers to be restored.

core/src/main/java/com/mapzen/android/graphics/model/BitmapMarkerFactory.java
```diff
@@ -0,0 +1,21 @@
+package com.mapzen.android.graphics.model;
+
+import com.mapzen.android.graphics.internal.StyleStringGenerator;
+
+/**
+ * Creates {@link BitmapMarker} objects.
+ */
+public class BitmapMarkerFactory {
```
tallytalwar: I might not be following the usage of the Factory here, but could this functionality not make sense in `MarkerManager` itself?

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
```diff
@@ -1,21 +1,39 @@
...
 
+import android.graphics.drawable.Drawable;
+
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Manages {@link BitmapMarker} instances on the map.
  */
 public class MarkerManager {
```
tallytalwar: I always get confused with this `Marker` name usage. I know its because multiple reasons including compatibility with how `Marker`->`MapData` usage was and how `BitmapMarker`->`Tangram::Marker` usage is. But can we try to keep the naming such that it distinguishes between sdk markers (tangram mapdata) and sdk bitmapmarkers (tangram markers)?

core/src/main/java/com/mapzen/android/graphics/GraphicsModule.java
```diff
@@ -34,4 +38,29 @@
...
+   */
+  @Provides @Singleton public StyleStringGenerator providesStyleStringGenerator() {
+    return new StyleStringGenerator();
+  }
+
+  /**
+   * Returns the object used to manager markers.
+   * {@link com.mapzen.tangram.Marker}.
+   */
+  @Provides @Singleton public MarkerManager providesMarkerManager(
```
tallytalwar: Does android sdk support multiple map instances (tangram android does!), if so does it make sense to not make `MarkerManager` singleton? Or will every map instance gets its own `MarkerManager`.

sarahlensing: The SDK doesn't support map instances, this is something I've had floating around in my head as a TODO. I created a ticket for it and will add support in a future PR https://github.com/mapzen/android/issues/436

core/src/main/java/com/mapzen/android/graphics/model/BitmapMarkerFactory.java
sarahlensing: It definitely could be within `MarkerManager` but the reason I pulled it out was to make testing easier. If I have a separate object in charge of creating the markers that I inject into `BitmapMarkerFactory`, I can easily have this factory object generate mocks and make assertions on them when testing:
https://github.com/mapzen/android/pull/433/files#diff-f00bf2e8f7b3f14b7d6bc8fbf88ddabfR40
https://github.com/mapzen/android/pull/433/files#diff-f00bf2e8f7b3f14b7d6bc8fbf88ddabfR57

tallytalwar: ohh yes definitely makes sense. 👍

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
sarahlensing: Good point. Updated to use `Collections.synchronizedList(restorableMarkers)`

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
sarahlensing: I use this method in two places- here when adding markers given `MarkerOptions` and then when restoring markers when I don't have options but instead only the `BitmapMarker` which is why I don't have it taking the `MarkerOptions` explicitly. I'll create a couple convenience methods on top of this one to handle configuring a marker given just the `MarkerOptions` and have it call through to this one. How does that sound? I could keep a mapping of `MarkerOptions` -> `BitmapMarker` and then have just one method to configure a marker but I'd prefer not to store this info if I can get away with it.

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
sarahlensing: I agree, the naming is not ideal. I'll update to `BitmapMarkerManager` for clarity

core/src/main/java/com/mapzen/android/graphics/model/MarkerOptions.java
sarahlensing: Updated to `BitmapMarkerOptions` and deprecated `MarkerOptions` !

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
tallytalwar: Yup that sounds fine to me.

core/src/main/java/com/mapzen/android/graphics/model/MarkerManager.java
```diff
@@ -48,11 +48,7 @@ public BitmapMarker addMarker(MarkerOptions markerOptions) {
     final Marker marker = mapController.addMarker();
     BitmapMarker bitmapMarker = bitmapMarkerFactory.createMarker(this, marker,
         styleStringGenerator);
-    configureMarker(bitmapMarker, markerOptions.getPosition(), markerOptions.getIconDrawable(),
-        markerOptions.getIcon(), markerOptions.getWidth(),
-        markerOptions.getHeight(), markerOptions.isInteractive(), markerOptions.getColorHex(),
-        markerOptions.getColorInt(), markerOptions.isVisible(), markerOptions.getDrawOrder(),
-        markerOptions.getUserData());
+    configureMarker(bitmapMarker, markerOptions);
```
tallytalwar: Yeah overloading this makes it more cleaner to me.

PR #448, opened 2017-09-08
Pull API Key

### LINKED ISSUE #444: HttpHandlers should pull api key from MapzenManager
Now that we provide a base abstract class for developers to extend if they need to provide custom headers or parameters, we should improve how the api key is configured. The handlers should pull the MapzenManager api key value and update it in response to changes in MapzenManager

sarahlensing:
### Overview
This improves how our base http handler classes set and update their knowledge of the Mapzen API key. The classes now automatically pull the API key from the `MapzenManager` and update it in response to changes in the `MapzenManager` class, making it so that developers who wish to add additional headers and parameters do not need to set or manually append the key

### Proposed Changes
- Creates a new listener interface, `ApiKeyChangeListener` for http handler classes to register with `MapzenManager` in order to receive updates when the API key is changed
- Updates search, routing, and tile request handling to automatically pull api key
- Removes `TmpHttpHandler` class now that we have ability to add custom headers
- Adds and updates test coverage

Closes #444


core/src/main/java/com/mapzen/android/core/MapzenManager.java
```diff
@@ -89,4 +97,23 @@ public void setApiKey(String apiKey) {
...
+   * Adds listener to list of managed callbacks so that it can be notified when API key changes
+   * occur.
+   * @param listenerReference
+   */
+  public void addApiKeyChangeListener(WeakReference<ApiKeyChangeListener> listenerReference) {
+    Collections.synchronizedList(listeners).add(listenerReference);
+  }
+
+  private void notifyListeners() {
+    for (WeakReference<ApiKeyChangeListener> weakReference : Collections.synchronizedList(
```
tallytalwar: Just for my knowledge sake :D, can you explain the need for `WeakReference` here?

mapzen-android-sdk/src/main/java/com/mapzen/android/routing/RouterInitializer.java
```diff
@@ -1,43 +1,37 @@
...
   /**
-   * Initialize the {@link MapzenRouter}'s api key in code and set it's
-   * {@link MapzenRouterHttpHandler}.
+   * Set the {@link MapzenRouter}'s {@link MapzenRouterHttpHandler}.
    * @param router
-   * @param apiKey
    */
-  public void initRouter(MapzenRouter router, String apiKey) {
-    requestHandler.turnByTurnHandler().setApiKey(apiKey);
+  public void initRouter(MapzenRouter router) {
```
tallytalwar: do you think it make sense to provide an overloaded method to provide a different apikey than the one set for `MapzenMap`?

core/src/main/java/com/mapzen/android/core/MapzenManager.java
sarahlensing: The APIs aren't balanced with `addApiKeyChangeListener` / `removeApiKeyChangeListener` , there is only  `addApiKeyChangeListener` (this is because the `finalize` method isn't a reliable place to cleanup resources so I only have `addApiKeyChangeListener`). Because of this, I use `WeakReference` to avoid listeners outliving their containing objects and leaking resources when added.

mapzen-android-sdk/src/main/java/com/mapzen/android/routing/RouterInitializer.java
sarahlensing: I think that since we moved to the unified API key, providing different keys for each Mapzen service shouldn't be necessary. I would say we wait to do this until a specific use case or request comes in for it.

PR #489, opened 2017-10-26
Deploy Stylesheets

### LINKED ISSUE #486: Deploy stylesheet artifacts

sarahlensing:
### Overview
This is an exciting change to the SDK. With this work, all house stylesheets are now deployed as separate artifacts which means that developers can optionally exclude or include them on an as-needed basis. 
```
compile 'com.mapzen:bubble-wrap:$VERSION_NUMBER'
compile 'com.mapzen:cinnabar:$VERSION_NUMBER'
compile 'com.mapzen:refill:$VERSION_NUMBER'
compile 'com.mapzen:tron:$VERSION_NUMBER'
compile 'com.mapzen:walkabout:$VERSION_NUMBER'
```
By default, the SDK includes the same four it always has (bubble-wrap, cinnabar, refill, and walkabout). But now tron is also deployed and can be included.

### Proposed Changes
- Moves all style submodules from `core/` to specific modules in `styles/`
- Updates `Tron` to latest release version which supports themes
- Sets up deployment for all new stylesheet modules (version numbers will be consistent with other module release numbers, ie. core, android sdk, and places api with 1.6.1 will result in corresponding stylesheet modules with version 1.6.1)
- Tron can be included as of the next release with this gradle syntax:
```
compile 'com.mapzen:mapzen-android-sdk:1.6.1'
compile 'com.mapzen:tron:1.6.1'
```
- Bubble-wrap, Cinnabar, Refill, or Walkabout can be excluded to thin app size in gradle:
```
compile ('com.mapzen:mapzen-android-sdk:1.6.1') {
  exclude group:'com.mapzen', module:'bubble-wrap'
  exclude group:'com.mapzen', module:'cinnabar'
  exclude group:'com.mapzen', module:'refill'
  exclude group:'com.mapzen', module:'walkabout'
}
```

Closes #486


matteblair:
From the description it looks like the scene file artifacts are going to use the same version number as the SDK, but I wonder if this would get confusing since the cartography assets already have their own versioning: https://mapzen.com/documentation/cartography/versioning/

sarahlensing:
I thought about keeping the scene file artifact version numbers consistent with the cartography releases but decided not to primarily because I wanted implementing developers to not have to think about which artifact version to use with any given version of the SDK. I think that most developers don't care which version of the cartography release they are using, they will just want the latest as quickly as possible. I also don't want to be tied directly to the cartography release versions so that if in the future we have to add code or files to one of these artifacts, it doesn't break the artifact release versioning.

